PR #34, opened 2019-08-30
Avoid using user's home directory by default

javierluraschi:
See https://github.com/rstudio/pins/issues/33.

CRAN requires packages to not use `~/pins` by default, the fix here is to:
- Remove boards registry, a `board.yml` file used to be stored in `~/pins`
- Avoid using the 'local' as the default board.

Some implications,
- Users will always have to register the boards they need, e.g. `board_register_local()`, this is quite easy to accomplish using the RStudio Connections pane and desirable for reproducibility.
- If users want to default to use the 'local' board for everything, they would need to set the `options(pins.board = "local")` setting.
- There new default board is a 'temp' board that stores data in the temporary R folder.
- RStudio now suggests using `keyring` if the `GITHUB_PAT` is not set, since no persistent board storage is available.


jjallaire:
I have written a bunch of packages that do save user data using standard system locations for storing cached data. How about if local uses rappdirs? (https://cran.r-project.org/web/packages/rappdirs/index.html)

javierluraschi:
@jjallaire `rappdirs` does not save data, correct? It reads from standard locations but does not seem to store data at all.

CRAN pointed out https://cran.r-project.org/web/packages/policies.html:

> Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.
>
> Limited exceptions may be allowed in interactive sessions if the package obtains confirmation from the user.

I think the challenge here is that `pins` is not requesting explicit confirmation from users to store data in the `~/.pins` folder.

I'm not questioning if this is a useful policy to have, but it happens to be one of their policies.

For users that are trying out pins and why not, might be a good idea to default to the temp folder to avoid using space, once they understand the package they can use `board_register_local()` or simply click "connect" from the RStudio Connections pane.

jjallaire:
rappdirs provides you the location of standard directories (e.g. the cache directory, the settings directory, etc.). It is up to you what you do with those directories (including writing to them). For example, sparklyr writes Spark installations to those directories. There is almost no point to reading those directories if you weren't the app that wrote to them in the first place, so we can assume that all of the packages that depend on rappdirs are writing data there.

javierluraschi:
Right, so I looked at a few packages that use `rappdirs`:

- `sparklyr`: Tests are not run in CRAN since they would be too heavy for CRAN to handle.
- `keyring`: An `rappdirs` path is only used in the file-base store and the tests override the location to use the temp folder, see [test-file.R#L8](https://github.com/r-lib/keyring/blob/a32781c4e8cb67f5b19b83ad33fdb80ddfad3615/tests/testthat/test-file.R#L8)
- `learnr`: Seems also to use `rappdirs` for a subset of the functionality, and the test overrides to use the temp folder as well, see [test-storage.R#L10](https://github.com/rstudio/learnr/blob/d762375b2737ddb828736318f645c8b408c65e30/tests/testthat/test-storage.R#L10).
- `hoarder`: This one actually advertises as a cache, which uses `rappdirs`; however, the tests and examples also override with a tempfile(), see [hoard_client.R#L119](https://github.com/ropensci/hoardr/blob/master/R/hoard_client.R#L119).

In the `pins` CRAN submission, we also overrode tests and examples with a temp folder for `pins`; in fact, all tests passed and the package got published to CRAN yesterday; however, it got manually inspected today and got taken down from CRAN, so I'm not sure that a `tempdir()` workaround will work for pins, even is we were to use `rappdirs`. I actually got pointed out the specific line in the sources that needs to be avoided.

Anyways, with this PR, users can now run `options(pins.board = "local")` to opt-in to make the local board the default board, as opposed to the temp folder.

Alternatively, having to write `board_register_local()`, makes it explicit that you are using a local cache and forces you to opt-in if you want a resource cached "permanently" vs your current session. The connection pane makes it really easy to connect back to boards and serves as the global registry the `pins` package was using.

In the spirit of getting this back in CRAN, I'm planning to take this change... but also happy to brainstorm other approaches we could consider.

README.Rmd
```diff
@@ -94,7 +94,7 @@ Finally, you can also **share** resources with other R sessions and other users
 To share with other R sessions, you can use a local board which stores pins in a shared path, usually `~/.pins`:
 
 ```r
-board_register_local(path = "~/pins")
+board_register_local(cache = "~/pins")
```
jjallaire: The docs say `~/.pins` but the code says `~/pins`

jjallaire:
Okay, hard to avoid this change if the code is being inspected line by line. Still, it seems like if the OS makes a user data cache folder available it would be great if we could use it! The automatic caching of datasets is definitely something I'd love for us to get back but I guess we'll have to think about whether there is an acceptable means of doing this.

PR #42, opened 2019-09-03
Improve readme

hadley:
I think my main contribution is to make it more clear that there are two parallel uses of `pin()`: for remote resources and for local objects.


javierluraschi:
Thank you, Hadley!

PR #112, opened 2019-11-12
Switch from Kaggle -> Azure for azure documentation

jduckles:
Fixing up docs, looks like you adapted the Kaggle vignette for Azure. Found two cases where Kaggle was referred to, switched them to Azure.


javierluraschi:
Thank you, @jduckles!

PR #116, opened 2019-11-13
Pass through account to wrapper for Azure

jduckles:
When testing `pins::board_register_azure()` it seems the wrapper wasn't passing the account string through. I think this proposed PR will fix that (untested).

Example of what is/isn't working for me:
image


javierluraschi:
Great catch, thank you!

PR #165, opened 2020-01-31
Allow user to specify RStudio Connect Front Matter

### LINKED ISSUE #159: Allow users to modify RStudio Connect index.html file
We're releasing some wrappers around pins for internal use, and we wanted to change the material which is printed in the Rstudio Connect gui, so that it would recommend our wrappers rather than `pin_get`. 

I did this in a fairly simple way by allowing the user to point to a different index.html file  which I'm happy to submit in a pull request if it's helpful: https://github.com/GShotwell/pins/blob/master/R/board_rsconnect_bundle.R#L4-L9

Another option for solving this problem might include some kind of pins/Rmarkdown hybrid (pindown) whereby the Rmarkdown was rendered as the front matter for the pin. This is probably more use friendly, but also a bit more complicated.

GShotwell:
This resolves #159 by allowing the user to specify a path to their own index.html file. This is a bit of a workaround, but we found it very helpful when developing custom pin functions in our internal package.

### DIFF
--- a/R/board_rsconnect_bundle.R
+++ b/R/board_rsconnect_bundle.R
@@ -1,13 +1,13 @@
 
+add_user_html <- function(dir, path = getOption("RSCONNECT_HTML_PATH")) {
+  if (path != "") {
+    file.copy(path, file.path(dir, "index.html"), overwrite = TRUE)
+  }
+}
+
 rsconnect_bundle_template_html <- function(temp_dir, template, value) {
   html_file <- file.path(temp_dir, "index.html")
-  user_file <- getOption("RSCONNECT_HTML_PATH")
-  if (!is.null(user_file)) {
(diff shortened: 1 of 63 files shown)


javierluraschi:
This looks great!

Would it help to open a feature request for something like a `pin-template`? As in, maybe we should allow creating a secret pin named `pin-template` in different boards which defines the base HTML to use when creating a pin...

@kevinykuo is also interested in this, as in, a way to create static websites in S3 when a pin is created...

Yeah, opening https://github.com/rstudio/pins/issues/166, feel free to add some thoughts there as well.

But yes, this PR looks good! -- Currently, the tests rely on a lot of cloud secrets so community PR's need to be tested separately, running tests here:

https://ci.appveyor.com/project/rstudio/pins/builds/30496932

Will merge afterwards.

PR #307, opened 2020-11-04
Port fix for rstudio/pins/pull/306

javierluraschi:
Just porting fix from https://github.com/rstudio/pins/pull/306/


jspiewak:
@javierluraschi just want to clarify, if the pin is being *updated* you don't delete, but if the pin is being *created* you still do?
Because I definitely ran into issues (early on) where I would have to go to the Connect UI and manually delete a partially created pin that failed to fully upload.

PR #311, opened 2020-11-07
fix #310 : test if path contains query before attempt to download data.txt

### LINKED ISSUE #310: pin() fails when a URL contains query parameters
Below is a reprex that tries to pin https://en.wikipedia.org/w/index.php?search=&title=Special%3ASearch&go=Go .
In the example, pinfails because of a malformed `data.txt`.
I took a look at its content, and found that `data.txt` is not a YAML summarizing the cache, but is a caching HTML document.

``` r
temp_dir <- tempdir()
pins::board_register_local(name = "local", cache = temp_dir)
xml2::read_html(pins::pin("https://en.wikipedia.org/w/index.php?search=&title=Special%3ASearch&go=Go"))
#> Error in yaml.load(string, error.label = error.label, ...): (/tmp/RtmpqTJQcQ/local/index_php_search_title_Special_3ASearch_go_Go/data.txt) Scanner error: mapping values are not allowed in this context at line 7, column 740
readr::read_lines(file.path(temp_dir, "local/index_php_search_title_Special_3ASearch_go_Go/data.txt"))[1:5]
#> [1] "<!DOCTYPE html>"                                     
#> [2] "<html class=\"client-nojs\" lang=\"en\" dir=\"ltr\">"
#> [3] "<head>"                                              
#> [4] "<meta charset=\"UTF-8\"/>"                           
#> [5] "<title>Search - Wikipedia</title>"
```

<sup>Created on 2020-11-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

atusy:
I added a test if URL contains query parameters before attempt to download data.txt to enable public access to boards like rsconnect.

Please see #310 for details.

### DIFF
--- a/R/pin_extensions.R
+++ b/R/pin_extensions.R
@@ -28,7 +28,7 @@ board_pin_store <- function(board, path, name, description, type, metadata, extr
   dir.create(store_path)
   on.exit(unlink(store_path, recursive = TRUE))
 
-  if (length(path) == 1 && grepl("^http", path) && !grepl("\\.[a-z]{2,4}$", path) && getOption("pins.search.datatxt", TRUE)) {
+  if (length(path) == 1 && grepl("^http", path) && !grepl("\\.[a-z]{2,4}$", path) && !grepl("\\?", path) && getOption("pins.search.datatxt", TRUE)) {
     # attempt to download data.txt to enable public access to boards like rsconnect
     datatxt_path <- file.path(path, "data.txt")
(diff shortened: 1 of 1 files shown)


javierluraschi:
Thanks for the PR, followed up with this https://github.com/rstudio/pins/compare/bugfix/pins-get-310?expand=1 -- We need to also clear the `data.txt` file otherwise other code paths might use it and break. That said, I really appreciate the interest to contribute here!

atusy:
Thanks for the follow up!

PR #345, opened 2020-12-21
Adding data structure validation to retrieved data.txt index

afshinmhCLG:
**Issue:** 
Currently **board_datatxt.R** assumes an index (e.g. data.txt) retrieved is valid so long as `httr::http_error(response) ` is FALSE. 

**Impact:** 
A corrupt data index or an API where missingness is embedded within the body of the response can break the workflow. As an example, I am extending pins to have a [labkey](https://www.labkey.org/home/project-begin.view?) board. Labkey [R API](https://cran.r-project.org/web/packages/Rlabkey/index.html) returns content errors within the body of the response. For example, if data.txt doesn't exist, `httr::http_error(response) ` still remains FALSE, while the body of the response would contain an error message informing about the missing file.

**Implemented solution:** 
In addition to `httr::http_error(response) `, content is evaluated to ensure `board_manifest_get()` and `board_manifest_load()` return a list structure containing element "path".  This requirement is currently **implicit in the code** and the implemented solution makes it explicit.


afshinmhCLG:
I'll try to address the test failure and reopen the PR

PR #351, opened 2021-01-19
Access type should be added to pin creation in rsconnect. Add test.

scottmmjackson:
A bug was reported through rstudio connect support that pins were not being created with the requested access_type, and instead would be created with the `acl` access_type.

New content created using the `pin` function crafts its API call here: (note the lack of an `access_type` argument passed)

https://github.com/rstudio/pins/blob/master/R/board_rsconnect.R#L120

Updating a pin using the `pin` function calls the API here:

https://github.com/rstudio/pins/blob/master/R/board_rsconnect.R#L151

This code adds the `access_type` argument and additionally updates tests such that we can prove we've fixed the bug.

### DIFF
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -123,7 +123,8 @@ board_pin_create.rsconnect <- function(board, path, name, metadata, code = NULL,
                                     app_mode = "static",
                                     content_category = "pin",
                                     name = name,
-                                    description = board_metadata_to_text(metadata, metadata$description)
+                                    description = board_metadata_to_text(metadata, metadata$description),
+                                    access_type = access_type
                                   ))
       if (!is.null(content$error)) {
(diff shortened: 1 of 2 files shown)


hadley:
I'm pretty sure this now fixed, but I don't quite have the right infrastructure to test it yet. Test will look something like this:

```R
test_that("can control acess type", {
  board <- board_rsconnect_test()
  board %>% pin_write(1:3, "test-x", access_type = "logged_in")
  withr::defer(pin_delete(board, "test-x"))
})  
```

hadley:
Also fixed another subtle bug — `access_type` was respected when the pin was created, but it wasn't updated after.

And then fixed another even subtler bug that fixing that subtle bug revealed — shouldn't change `access_type` unless it's explicitly specified.

PR #352, opened 2021-01-19
Standardise github action

hadley:


hadley:
Fixes #353

.github/.gitignore
```diff
@@ -0,0 +1 @@
+*.html
```
scottmmjackson, line 1: Is this to exclude rendered output of Rmds?

scottmmjackson (changes requested):
This looks good. It makes me think that maybe a `HACKING.md`/`CONTRIBUTING.md` containing instructions (i.e.: the list of environment variables you should set) for configuring a full test run would be helpful.

scottmmjackson (approved):

.github/.gitignore
hadley: Yeah, if you have any `.md` files in that directory (it's automatically added by some helper in usethis)

hadley:
Yeah, that would be nice to do one day 😄 At least the skip messages now tell you what you need to set.

PR #358, opened 2021-01-29
Handle files in directories in pin_download_one(), closes #357

### LINKED ISSUE #357: curl/httr error in get_pin() / pin_download_one() when file is inside a directory
In brief: when a pin contains a directory with files in it, `get_pin()` fails at the `pin_download_one()` call.

An end-to-end reprex would be hard to share since creating the pin would require the method I am implementing in a package: I am creating a custom pin method following the [Extending pins vignette](https://pins.rstudio.com/articles/pins-extending.html), specifically for Arrow datasets which require the data files to sit in a separate directory.

Below is a limited reprex that uses an [existing pin](https://github.com/petrbouchal/storage-public/tree/main/arrow_ds) created to have files nested in a directory:

``` r
library(pins)
github_public <- pins::board_register_github(repo = "petrbouchal/storage-public", branch = "main")
x <- pin_get("arrow_ds", board = github_public, cache = F)
#> Error in curl::curl_fetch_disk(url, x$path, handle = handle): Failed to open file /var/folders/c8/pj33jytj233g8vr0tw4b2h7m0000gn/T//Rtmp51LbdF/file142807b0d2611/arrow_files/part-0.parquet.
```

<sup>Created on 2021-01-29 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

The error occurs  in: https://github.com/rstudio/pins/blob/305bfbc2adfd8089151fa4bd3654c1fed2d3b705/R/pin_download.R#L99-L104

`destination_path` is correctly defined, but if it points to a file in a subdirectory, the `httr::GET()` call fails because that subdirectory has not been created.

I will submit my local fix as a PR for you to consider.

possibly related to #348

petrbouchal:
The PR ensures that before downloading a file in `pin_download_one()`, all directories contained in the destination path are created on disk inside the temp dir before the download is attempted.

This handles cases where a pin contains files in directories, which currently causes an error.

See issue #357 for details.﻿

Reprex to verify this, relying on an existing pin with a file inside a directory. 

``` r
library(pins)
github_public <- pins::board_register_github(repo = "petrbouchal/storage-public", branch = "main")
x <- pin_get("arrow_ds", board = github_public, cache = F)
```

<sup>Created on 2021-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/pin_download.R
+++ b/R/pin_download.R
@@ -97,6 +97,7 @@ pin_download_one <- function(path,
 
         if (remove_query) download_name <- strsplit(download_name, "\\?")[[1]][1]
         destination_path <- file.path(temp_path, download_name)
+        dir.create(dirname(destination_path), recursive = T, showWarnings = F)
         pin_log("Downloading ", path, " to ", destination_path)
         details$something_changed <- TRUE
 


hadley:
Closing this since we're generally moving away from this code path.

Issue #359, opened 2021-02-03
Problem in datatxt_refresh_index when subpath is specified


lenafm:
I could be completely wrong with this but I think the `httr::GET` part of `datatxt_refresh_index` fails when the subpath is not null (in my case this is because boards are 'folders' inside an AWS S3 bucket):

```
response <- httr::GET(index_url,
                                   httr::write_disk(temp_index, overwrite = TRUE),
                                   board_datatxt_headers(board, "data.txt"))
```

If I understand the code correctly, this fails because `board_datatxt_headers(board, "data.txt")` is only passed `"data.txt"` but it should have the subpath attached to it. 

In `datatxt_update_index` the `httr::GET` part works, because the index file does have the subpath attached to it:

```
response <- httr::GET(file.path(board$url, index_file_get),
                                   board_datatxt_headers(board, index_file_get))
```
where `index_file_get <- file_path_null(board$subpath, "data.txt")`.

hadley:
I'm going to rewrite `board_s3()` from scratch in the near future, so this problem will be fixed along the way.

PR #360, opened 2021-02-05
Pass correct index_file to board_datatxt_headers

lenafm:
In `datatxt_refresh_index` the index file passed to `board_datatxt_headers` in `httr::GET(..)` was `"data.txt"` but - in order to work with boards that have a subpath - the index file passed should be `index_file_get <- file_path_null(board$subpath, "data.txt")`.

This is fixed in this PR. For obvious reasons I couldn't run the tests that are skipped automatically because I wasn't able to connect to the various remote platforms. All other tests passed.

Fixes issue #359

### DIFF
--- a/R/board_datatxt.R
+++ b/R/board_datatxt.R
@@ -9,16 +9,17 @@ datatxt_refresh_index <- function(board) {
   if (is.null(board$url)) stop("Invalid 'url' in '", board$name, "' board.")
 
   index_file <- "data.txt"
-  if (identical(board$index_randomize, TRUE)) {
-    index_file <- paste0(index_file, "?rand=", stats::runif(1) * 10^8)
-  }
-
   index_url <- file_path_null(board$url, board$subpath, index_file)
+
+  index_file_get <- file_path_null(board$subpath, "data.txt")
+  if (identical(board$index_randomize, TRUE)) {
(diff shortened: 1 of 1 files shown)


R/board_datatxt.R
```diff
@@ -9,16 +9,17 @@ datatxt_refresh_index <- function(board) {
   if (is.null(board$url)) stop("Invalid 'url' in '", board$name, "' board.")
 
   index_file <- "data.txt"
+  index_url <- file_path_null(board$url, board$subpath, index_file)
+
+  index_file_get <- file_path_null(board$subpath, "data.txt")
```
hadley, line 14:
```suggestion
  index_file_get <- file_path_null(board$subpath, index_file)
```
?

lenafm: Thanks @hadley, I have updated and commited your suggested change.

hadley:
Now that I understand the datatxt format, unfortunately I have determined that it is not a good fit for pins' needs. The problem is maintaining a single central list of pins — if two pins should be written at the same time, there is a substantial chance of a race condition so that only one pin will recorded in the site metadata. This means that I don't think it's worthwhile to patch up smaller bugs in (what is now) `legacy_datatxt()`.

PR #411, opened 2021-03-23
Add CI workflow for testing RSC

colearendt:
adds a CI workflow for running RStudio Connect


hadley:
Can you fill in the dots on connecting to the server from R? And creating new users etc?

colearendt:
Oh definitely. Yes, apologies for forgetting that part.

colearendt:
Alright @hadley the first user is created and has an API key defined. I'm not sure that creating additional users is super functional right now, but I would love to understand whether that would be helpful or not, so I can pass it along to the Connect team.

(To be clear, creating additional users is pretty easy. Creating API keys for those users to interact with the product as publishers is the very challenging part).

.github/workflows/rstudio-connect.yaml
```diff
@@ -0,0 +1,89 @@
...
+      # TODO: add real integration tests here
+      - name: run tests
+        shell: Rscript {0}
+        run: |
+          # this should read the env vars by default... but it is not
+          # I think that is a bug
+          # pins::board_register_rsconnect()
+
+          # reading the env vars manually
+          pins::board_register_rsconnect(server = Sys.getenv("CONNECT_SERVER"), key = Sys.getenv("CONNECT_API_KEY"))
```
hadley, line 83: What sets these env vars?

colearendt: A few lines up (two steps up) - `connectapi:::update_renviron_creds(host = key$host, api_key = key$api_key, prefix = "CONNECT")`

It writes the `.Renviron` file so the env vars can be picked up / used.

colearendt: Long term: I am planning to create a proper action for this set up here: https://github.com/rstudio/actions

hadley:
What's the best way to create another user? I want to be able to write tests around pins with the same name by different users

colearendt:
Does the other user need to "do things" / have a token / API key / etc.? I'll take a look at it tomorrow

hadley:
@colearendt yeah, I think so. Don't worry too much about it; this is still a big improvement over the current system.

colearendt:
Ok, I’ll take a quick look! I have an idea that I hadn’t tried before that could simplify it. Ultimately there are a few features that Connect could add to make this much easier.

colearendt:
Alright I have this working - had to switch to using PAM auth, which is a bit less than ideal. Basically, to create a user you have to:
- add the user to `users.txt` (`user` / `password`)
- call something like: `con$users_create("test", "test@test.com")` as an admin / the first user (I think... to set the email for users. It may be that you can get around this)
- Do something like this to get an API key for a user. If wrapping this in an internal function inside of `connectapi` would make things easier, I'm happy to do that!
```
client <- HackyConnect$new(host = con$host, api_key = NULL)
client$login("test", "test")
user_key <- client$POST(path = "keys", body = list(name = "mykey"))
```

Ultimately, I think Connect needs to make this story easier. Initializing users is painful today

hadley:
@colearendt thanks for all your work on this, and I'm finally coming back to it with the intention of getting it merged so I can make sure we can be very clear about which versions of RSC pins supports. It looks like something in connectapi has changed since you last worked on this, so would you mind taking another look? Answer to your other questions below:

* Do you need a way to run this same pattern locally? — Yes, ideally, but my hacky testing against `connect.rstudioservices.com` is fine for the short term
* Would you like for me to go through this and initialize all of the users in users.txt by calling these commands? — I think just creating two imaginary users (e.g. "susan" and "michael") would be fine
*  Where should I store the keys? In an object inside of the R CMD check block? — I think saving an rds file into the root directory would be fine; then I can use that in the tests.

hadley:
Note to self: the goal is be to get to a point where I can interact simultaneously with two different users in the tests, so I can make sure the cross-user stuff works, e.g.:

```R
susan <- board_rsconnect(account = "susan", ...)
derek <- board_rsconnect(account = "derek", ...)

susan %>% pin_write(1:10, "test1")
derek %>% pin_read("susan/test1")
```

colearendt:
@hadley Apologies for the delay! I believe this is fixed (although CI will tell). I saved the text version of the API keys into an RDS file called `creds.rds` under a list with keys `susan_key` and `derek_key`:

https://github.com/rstudio/pins/blob/07cdc3a04e59e2460e2e3e71a78478b937e2aec9/.github/workflows/rstudio-connect.yaml#L77-L84

I'll work on a follow-up later that pulls this into a separate tool (like a bash script or something more portable) for running locally.

hadley:
Thanks for all your work on this @colearendt!

PR #420, opened 2021-04-01
removes broken python package

sellorm:
This particular python package doesn't work properly, was never published and is one of 4 python pins implementations in the wild.

I suggest we remove it to avoid confusion.


hadley:
Thanks!

PR #422, opened 2021-04-09
build dev site

edavidaja:
before pins 1.0 is available on CRAN it may confuse current connect users that the pins docs and connect docs are out of sync (it confused me).


hadley:
Thanks! I also rolled the main pins site back.

PR #424, opened 2021-04-13
Change .htm to .html in Vignettes

jthomasmock:
There are 3 broken links in the vignettes/examples due to `.htm` endings rather than `.html`. This PR corrects the file ending so that they can link to the correct pieces of content.


hadley:
Thanks!

PR #439, opened 2021-04-27
Use the download-all endpoint to download data from competitions.

### LINKED ISSUE #348: pin_get() fails with kaggle dataset that contains directory
The code and the error follows:

``` r
pins::board_register_kaggle(token = "~/kaggle.json")
file <- pins::pin_get("c/birdsong-recognition", board = "kaggle", extract = TRUE)
#> Error in pin_download_one(p, ...): Client error: (404) Not Found. Failed to download remote file: https://www.kaggle.com/api/v1/competitions/data/download/birdsong-recognition/example_test_audio/ORANGE-7-CAP_20190606_093000.pt623.mp3
```

<sup>Created on 2021-01-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I can't tell if this occours due to some pin issue or due to kaggle's api issue. Google didn't help me, though =(

dfalbel:
This should fix #348 by using the `download-all` endpoint instead of `download`. 

The `download-all` endpoint downloads a zip file containing all files from the competition, thus we don't need to list all files and download them one by one. We no longer need to change the `extract` argument too.

### DIFF
--- a/R/board_kaggle.R
+++ b/R/board_kaggle.R
@@ -273,13 +273,12 @@ kaggle_competition_files <- function(board, name) {
 #' @export
 board_pin_get.pins_board_kaggle <- function(board, name, extract = NULL, version = NULL, ...) {
   if (grepl("^c/", name)) {
-    competition_files <- name <- gsub("^c/", "", name)
+    name <- gsub("^c/", "", name)
     if (!grepl("/", name)) {
       files <- kaggle_competition_files(board, name)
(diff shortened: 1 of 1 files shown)


hadley:
I'm going to tackle this a slightly different way in `board_kaggle()`; I'm pretty sure it's better to iterate over the files individually so we can better make use of the cache if the data hasn't changed.

PR #441, opened 2021-04-29
Debug CI failures

hadley:


hadley:
Thanks to @gaborcsardi's investigation the bug comes from writing Latin1; need to re-encoded all character elements with `enc2utf8()`

hadley:
```R
to_utf8 <- function(x) {
  if (is.list(x)) {
    names(x) <- enc2utf8(x)
    lapply(x, to_utf8)
  } else if (is.character(x)) {
    enc2utf8(x)
  } else {
    x
  }
}
```

hadley:
Root cause: https://github.com/viking/r-yaml/issues/90

juniperlsimonis:
> 
> 
> Root cause: [viking/r-yaml#90](https://github.com/viking/r-yaml/issues/90)

thank you for linking this, having a similar issue with a package i'm working on and trying to track down why it was crashing

PR #471, opened 2021-07-15
Fix the URL on servers that have hostname + path

### LINKED ISSUE #469: Error while reading pin metadata on RStudio Connect
I am using the current dev version of the package and can use `pin_write()` and `pin_list()` just fine, but am having trouble with `pin_read()` because this line in `pin_meta.pins_board_rsconnect` gives a 404:

https://github.com/rstudio/pins/blob/322b3df464202ef3a4726cded3f1a549cefef11c/R/board_rsconnect.R#L249

``` r
library(pins)

b <- board_rsconnect("envvar")
b %>% pin_read("julia.silge/biv_svm")
#> Error in rsc_check_status(req): Not Found (HTTP 404).
```

<sup>Created on 2021-06-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

This happens for all pins I try to read. If I try to read again, I get a different error because `data.txt` has been saved with HTML in it.

```
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.14.0 (Ubuntu)</center>
</body>
</html>
```

andrie:
This attempts to fix #469, although some more work might be required, since the tests don't all pass locally.

The root cause is that the function `board_rsconnect()` constructs the server URL from `hostname` + `__api__`.  However, this ignores the fact that the full domain name can contain `hostname` + `path` + `__api__`.

For example, if the real server is at `colorado.rstudio.com/rsc`, then the API endpoint is at `colorado.rstudio.com/rsc/__api__`, but the current code base ignores the `/rsc` path.

One solution to this is to use `rsconnect::serverInfo()` to construct the URL, like this:

```r
url <- rsconnect::serverInfo(name = server)$url
```

instead of the current:

```r
url <- paste0(server, "/__api__/") # remember to delete this line
```

Here is a reprex:

```
board <- board_rsconnect(server = "https://colorado.rstudio.com/rsc")
pin_read(board, "andrie/mtcars-test")
```

### DIFF
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -71,7 +71,8 @@ board_rsconnect <- function(
     server <- server %||% envvar_get("CONNECT_SERVER") %||% abort("`server` must be supplied")
     server_name <- httr::parse_url(server)$hostname
     # account determined below
-    url <- paste0(server, "/__api__/")
+    url <- paste0(server, "/__api__/") # remember to delete this line
+    url <- rsconnect::serverInfo(name = server)$url
 
     key <- key %||% envvar_get("CONNECT_API_KEY") %||% abort("`key` must be supplied")
     account_info <- NULL
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -597,6 +597,7 @@ rsc_path <- function(board, path) {
 
 
 rsc_ensure_path_url <- function(board, url) {
+  if (length(url) == 0) return(url)
   server_name <- dirname(board$server_name)
   full_name <- board$server_name
   if (!grepl(full_name, url)) {
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -600,7 +600,7 @@ rsc_ensure_path_url <- function(board, url) {
   if (length(url) == 0) return(url)
   server_name <- dirname(board$server_name)
   full_name <- board$server_name
-  if (!grepl(full_name, url)) {
+  if (!grepl(full_name, url, fixed = TRUE)) {
     url <- sub(server_name, full_name, url)
   }
   url
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -451,6 +451,16 @@ board_pin_find.pins_board_rsconnect <- function(board,
 
 # Content -----------------------------------------------------------------
 
+rsc_ensure_path_url <- function(board, url) {
+  if (length(url) == 0) return(url)
+  server_name <- dirname(board$server_name)
+  full_name <- board$server_name
+  if (!grepl(full_name, url, fixed = TRUE)) {
+    url <- sub(server_name, full_name, url)
+  }
+  url
+}
+
 rsc_content_find <- function(board, name, version = NULL, warn = TRUE) {
   name <- rsc_parse_name(name)
 
@@ -492,6 +502,8 @@ rsc_content_find <- function(board, name, version = NULL, warn = TRUE) {
     }
     selected <- json[[name$owner %in% owner_names]]
   }
+
+  selected$content_url <- rsc_ensure_path_url(board, selected$content_url)
 
   content <- list(
     guid = selected$guid,
@@ -596,16 +608,6 @@ rsc_path <- function(board, path) {
 }
 
 
-rsc_ensure_path_url <- function(board, url) {
-  if (length(url) == 0) return(url)
-  server_name <- dirname(board$server_name)
-  full_name <- board$server_name
-  if (!grepl(full_name, url, fixed = TRUE)) {
-    url <- sub(server_name, full_name, url)
-  }
-  url
-}
-
 rsc_GET <- function(board, path, query = NULL, ...) {
   path <- rsc_path(board, path)
   auth <- rsc_auth(board, path, "GET", NULL)
@@ -617,12 +619,7 @@ rsc_GET <- function(board, path, query = NULL, ...) {
     ...
   )
   rsc_check_status(req)
-  content <- httr::content(req)
-  if (length(content) && is.list(content) && is.list(content[[1]])) {
-    # ensure the full server name is in the URL
-    content[[1]][["content_url"]] <- rsc_ensure_path_url(board, content[[1]][["content_url"]])
-  }
-  content
+  httr::content(req)
 }
 
 rsc_download <- function(board, content_url, dest_path, name) {
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -69,10 +69,13 @@ board_rsconnect <- function(
   auth <- check_auth(auth)
   if (auth == "envvar") {
     server <- server %||% envvar_get("CONNECT_SERVER") %||% abort("`server` must be supplied")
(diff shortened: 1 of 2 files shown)


R/board_rsconnect.R
```diff
@@ -71,7 +71,8 @@ board_rsconnect <- function(
     server <- server %||% envvar_get("CONNECT_SERVER") %||% abort("`server` must be supplied")
```
hadley, line 71: I wonder if we should instead use `CONNECT_URL`, which we'd expect to contain the full path, not just the hostname as `SERVER` suggests.

hadley: Regardless, since I seem to keep messing this up, it also needs a decent comment explaining the problem. And maybe it should perform a simple request (maybe to the version endpoint?) so error early if the server is misconfigured. 

I don't think your approach will work because there's no guarantee that the server will be registered with rsconnect.

andrie: You're correct - my approach doesn't work for exactly the reason you mention.  I commented on why the current approach doesn't work at https://github.com/rstudio/pins/issues/469#issuecomment-881016721.  I'll edit this PR comment to include the root cause of the problem.

andrie: I have pushed another attempt at fixing this:
* Without using `rsconnect`
* I inspect the URL at pin write time as well as pin read time
* I created a helper function `rsc_ensure_path_url()` that does the work
*

juliasilge:
This looks like it is working for me now! 🤩

R/board_rsconnect.R
hadley: Why can't that just be done on board creation?

andrie: That's a great question.   I think you're correct that doing this at board creation time is going to be sufficient.

tl'dr:  I'll remove the `rsc_ensure_path_url()` in my next commit.

I understand more about how the package works now, and some of my approach was informed by an intermediate stage where "broken" pins were published on Connect.  It seems that the pin itself (maybe), as well as the local cache (certainly), contains the path to the pin.  This meant that while you have a local cache that contains an incorrect URL, my code was actually necessary.

Now that both I and @juliasilge have cleared our local cache, it seems that the much simpler code base will just work.  In other words, there was this intermediate stage where only a few people would be affected by a broken local cache URL.  Since my testing subjects were in that state, it affected how I thought the package works.

Anyway, inbound commit coming soon.

juliasilge:
With this last commit ⬆️ I can now `pin_write()` as well.

R/board_rsconnect.R
```diff
@@ -587,7 +592,18 @@ update_cache <- function(path, key, value) {
...
-  paste0("/", board_path, "/", path)
+  file.path(board_path, path, fsep = "/")
+}
+
+
+rsc_ensure_path_url <- function(board, url) {
+  if (length(url) == 0) return(url)
+  server_name <- dirname(board$server_name)
+  full_name <- board$server_name
+  if (!grepl(full_name, url)) {
```
hadley, line 603: Should probably use `fixed = TRUE` here to be safe

R/board_rsconnect.R
```diff
@@ -587,7 +592,18 @@ update_cache <- function(path, key, value) {
 
 rsc_path <- function(board, path) {
   board_path <- httr::parse_url(board$url)$path
-  paste0("/", board_path, "/", path)
+  file.path(board_path, path, fsep = "/")
```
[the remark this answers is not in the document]
andrie: This is necessary because sometimes there is a trailing `/` in `board_path` and sometimes not.

R/board_rsconnect.R
```diff
@@ -601,7 +617,12 @@ rsc_GET <- function(board, path, query = NULL, ...) {
     ...
   )
   rsc_check_status(req)
-  httr::content(req)
+  content <- httr::content(req)
+  if (length(content) && is.list(content) && is.list(content[[1]])) {
```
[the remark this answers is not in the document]
andrie: I think this is the right place, since the problem actually occurs in the line 620 in `httr::content(req)`.

Fixing it here fixes the problem when it occurs.  If we push the fix downstream, you will have to potentially fix in multiple places.

R/board_rsconnect.R
andrie: Fixed.

R/board_rsconnect.R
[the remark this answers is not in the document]
hadley: And what does `file.path()` do?

R/board_rsconnect.R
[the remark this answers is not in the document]
hadley: The currently code assumes that every response from the server has `content_url` in the first element of the response, which seems odd to me. 

If look at `board_rsconnect.R`, the only place I can see `content_url` being read is `rsc_content_find()` (which saves it in the cache), so can't we fix it there?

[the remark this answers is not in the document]
andrie: This pointer to `rsc_content_find()` is helpful.  I'll investigate.

R/board_rsconnect.R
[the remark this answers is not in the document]
andrie: Hmm.  I was wrong.  I thought `file.path()` does the "right" thing, but it doesn't.  Thanks for pointing this out.  Let me ponder on the best way of fixing this.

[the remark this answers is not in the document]
hadley: It looks like it will strip a trailing `/`, but only on windows: https://github.com/wch/r-source/blob/1cd91c2156bdcd8d5c3677b6b3aa00b298ee8085/src/main/paste.c#L424-L433. In general, I don't trust `file.path()` because of this sort of platform specific behaviour.

hadley:
I'll take it from here. Thanks for all your work on this!

hadley:
Isn't the root cause of the problem that the RSC api is returning the wrong url? i.e. the content url for `julia.silge/biv_svm` is given as `https://colorado.rstudio.com/content/971208fa-72aa-4a1d-ba00-7de2dd9a086a/` but it should actually be `https://colorado.rstudio.com/rsc/content/971208fa-72aa-4a1d-ba00-7de2dd9a086a/`?

hadley:
Argh, or did I get confused by the cache

andrie:
> Argh, or did I get confused by the cache

Yes, you have to clear out any cache that may still refer to old URLs.  Since I programmatically want to clear the cache regularly, without asking for confirmation, I defined locally:

```r
cache_prune <- function(days = 30, confirm = FALSE) {
  to_delete <- cache_old(days)
  if (length(to_delete) == 0) {
    pins_inform("No stale pins found")
    return(invisible())
  }

  size <- dir_size(to_delete)

  if (!confirm) {
    pins_inform("Delete {length(to_delete)} pin versions, freeing {format(size)} ?")
  }

  if (confirm || utils::menu(c("Yes", "No")) == 1) {
    fs::dir_delete(to_delete)
  }
}
```

hadley:
Thanks for working on this! I'm going to go with #477, but this helped me better understand the problem and catalysed me to fix it 😄

PR #477, opened 2021-08-04
Refactor server discovery

### LINKED ISSUE #469: Error while reading pin metadata on RStudio Connect

hadley:
Includes several debugging enhancements:

* Always prints url of server that you've connected to
* Perform an API request in board_rsconnect() to check that server works

Appears to fix #469, but I'm not sure why.

@juliasilge can you give this one a try please? 
@andrie this ended up looking rather different to your PR, but it seems to fix the problem, which surprises me.


andrie:
This works for me, and I also find it hard to reason about what changed 🤷 

One observation: we need to strip any trailing `/` in the `server` argument`.  Previously this wasn't necessary, since the `httr` machinery would remove any trailing `/`.

hadley:
Ok, I'll add a couple of explicit tests around handling urls with and without trailing slashes (which seems likely to force me to refactor the code a little) and then merge this PR.

juliasilge:
This works for me 🙌 

Related to @andrie's comments ⬆️ I did have a trailing slash in my URL earlier (my `.Renviron` had `CONNECT_SERVER = https://colorado.rstudio.com/rsc/` previously) and this worked with the old version of pins but was one of the sources of the problems I'd been experiencing.

PR #478, opened 2021-08-06
Implement board_azure()

hadley:


R/board_azure.R
```diff
@@ -0,0 +1,187 @@
...
+#' }
+#'
+#' \dontrun{
+#' # To create a board that you can write to, you'll need to supply one
+#' # of `key`, `token`, or `sas`
+#' board <- board_azure("pins", "private-data", sas = "...")
+#' board %>% pin_write(iris)
+#' }
+board_azure <- function(
+                        account,
```
hadley, line 28: @hongooi73 do you have any sense of whether specifying a Azure storage container with account + container name is a good idea? Or do you think people would prefer to just provide the whole url?

hongooi73: If it's just the choice between those 2 alternatives, I'd say providing the URL is better. This is because Azure actually has [multiple clouds](https://docs.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure), eg for the US govt cloud, the storage endpoint URL is of the form `https://{acct}.{service}.core.usgovcloudapi.net`.

However, I think the best solution is to leverage the existing framework in AzureStor, which should handle all of this under the hood. In particular, by calling the `storage_container` generic, you can support not just blobs, but also file and ADLSgen2 storage -- the latter is kind of the "next-generation" of blob storage, offering faster and more powerful access to blobs.

I'd do something like this:

```r
board_azure <- function(container, ..., versioned=TRUE, cache=NULL) {

  check_installed("AzureStor")

  # this will handle multiple ways to specify the container, as well as different services: file, blob, ADLS2
  if(!inherits(container, "storage_container")) {
    container <- storage_container(endpoint=container, ...)
  }

  cache <- cache %||% board_cache_path(paste0("azure-", hash(container$endpoint$url)))
  new_board_v1(...)
}
```
This means you also don't need to worry about Azure specifics like the key, SAS, token, etc -- that's AzureStor's problem. This way, you can just pass anything in `...` along and it can deal with them.

hongooi73: Similarly, everywhere that you call `list_blobs`, `upload_blob` etc I'd use the generic version: `list_storage_files`, `storage_upload`, ....

hadley: Oh good idea, thanks!

PR #502, opened 2021-09-06
compute Azure storage hash using container URL

### LINKED ISSUE #501: board_azure possibly incorrectly computing hash of URL
In the dev version, we have:

```r
board_azure <- function(container, n_processes = 10, versioned = TRUE, cache = NULL) {
  check_installed("AzureStor")

  cache <- cache %||% board_cache_path(paste0("azure-", hash(url)))
  ...
```
However, the URL of the storage container is actually in the component `container$endpoint$url`; the code as-is is computing the hash of the base function `url()`.

hongooi73:
fixes #501


hadley:
Thanks!

PR #507, opened 2021-09-10
Clarify board_rsconnect allows manual authentication

### LINKED ISSUE #508: Clarify manual authorization is allowed in board_rsconnect()
Amend  `board_rsconnect()` to mention that you can provide `key` and `server` directly for authorization. This is the first choice used by `auth="auto"`, not "envvar" as indicated in the documentation.

smingerson:
Mention manual authorization in `board_rsconnect()`. I don't know if it makes sense to change the function to accept a separate manual option or not. If you think so, I can add something like that. 

Closes #508

### DIFF
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -35,9 +35,11 @@
 #'
 #' @inheritParams new_board
 #' @inheritParams board_url
-#' @param auth There are three approaches to auth: use `"manual"` and provide `server` and `key`,
-#'   use `"envvars"`  `CONNECT_API_KEY` and `CONNECT_SERVER`, or the rsconnect package. The
-#'   default is `auto`, which will use the manually provided values if present, then
+#' @param auth There are three approaches to auth:
+#'  * Use `"manual"` along with arguments `server` and `key`.
+#'  * Use `"envvars"` with environment variables `CONNECT_API_KEY` and `CONNECT_SERVER`.
+#'  * Use the `"rsconnect"` package.
+#'  * The default is `auto`, which will use the manually provided values if present, then
 #'   environment variables if both are available, and rsconnect if not.
 #' @param server For `auth = "manual"` or `auth = 'envvar'`, the full url to the server,
 #'   like `http://server.rstudio.com/rsc` or `https://connect.rstudio.com/`.
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -35,12 +35,12 @@
 #'
 #' @inheritParams new_board
 #' @inheritParams board_url
-#' @param auth There are two approaches to auth: you can either use `"envvars"`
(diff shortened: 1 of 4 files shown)


hadley:
Yeah, I was actually wondering about this the other day. So if someone else had the same thought, I think it's a good sign it's worth doing 😄. I think "auto" auth would select "manual" if both `server` and `key` were supplied, so most of the changes would be to `check_auth()` and then the docs for `board_rsconnect()`. Is that clear enough for you to have a go at it?

smingerson:
Yes it is, I'll take a crack at it in the next couple days.

R/board_rsconnect.R
```diff
@@ -35,15 +35,15 @@
...
 #' @inheritParams new_board
 #' @inheritParams board_url
-#' @param auth There are two approaches to auth: you can either use `"envvars"`
-#'   `CONNECT_API_KEY` and `CONNECT_SERVER` or the rsconnect package. The
-#'   default is `auto`, which will use the environment variables if both are
-#'   available, and rsconnect if not.
-#' @param server For `auth = "envvar"` the full url to the server, like
-#'   `http://server.rstudio.com/rsc` or `https://connect.rstudio.com/`.
+#' @param auth There are three approaches to auth: use `"manual"` and provide `server` and `key`,
+#'   use `"envvars"`  `CONNECT_API_KEY` and `CONNECT_SERVER`, or the rsconnect package. The
```
hadley, line 39:
```suggestion
#'   use `"envvars"`  `CONNECT_API_KEY` and `CONNECT_SERVER`, or use the "rsconnect" package. The
```

I wonder if it might be clearer to switch this to a bullet list?

hadley (commented):
Looks good; just a couple of small comments.

smingerson:
I removed `rsc_server_envvar()`, replacing it with `rsc_server_manual()`, as that can be used as `rsc_server_manual(server, key)` or `rsc_server_manual(envvar_get("CONNECT_SERVER"), envvar_get("CONNECT_API_KEY")`. Let me know if that's satisfactory, I can put back `rsc_server_envvar()` back in too if you would prefer.

R/board_rsconnect_server.R
```diff
@@ -27,12 +32,12 @@ check_auth <- function(auth = c("auto", "envvar", "rsconnect")) {
   }
 }
 
-rsc_server_envvar <- function(server = NULL, key = NULL) {
-  url <- server %||% envvar_get("CONNECT_SERVER") %||% abort("`server` must be supplied")
-  url <- rsc_normalize_server_url(url)
+rsc_server_manual <- function(server, key) {
+  url <- server %||% abort("`server` must be supplied")
+  url <- rsc_normalize_server_url(server)
```
hadley, line 37:
```suggestion
  url <- rsc_normalize_server_url(url)
```

I think?

hadley (commented):
Exactly what I was imagining, thanks! Just one last question below.

Issue #482, opened 2021-08-18
Improve set up for azure/s3 test data


hadley:
* Use RStudio accounts so that others can access if needed
* Figure out how to get [paws](https://github.com/paws-r/paws/blob/main/docs/credentials.md#get-credentials-from-the-aws-shared-credentials-file) working with SSO (https://github.com/paws-r/paws/issues/457)
* Document

hadley:
And think about `board_ms365()`.

PR #513, opened 2021-09-13
Onedrive/Sharepoint board

hongooi73:
Implements boards in Onedrive/Sharepoint Online, using Microsoft365R as the backend. These use the same underlying API.

There are 3 top-level functions `board_onedrive`, `board_sharepoint` and `board_ms365`. All of these are the same, I've duplicated them because "onedrive" and "sharepoint" are the well-known names whereas "ms365" is rather anonymous. The lower-level code uses "ms365" as the prefix.

Pass an object of class `ms_drive` (a personal or business Onedrive, or a Sharepoint document library) to create a board:

```r
library(Microsoft365R)
od <- get_personal_onedrive()
board_od <- board_onedrive(od, "path/to/boardname1")

odb <- get_business_onedrive()
board_odb <- board_onedrive(od, "path/to/boardname2")

doclib <- get_sharepoint_site("mysitename")$get_drive("doclibraryname")
board_sp <- board_sharepoint(doclib, "path/to/boardname3")
```

Basic uploading, downloading and metadata seems to be functional, will try to complete this over the next week.


hongooi73:
Deleting files in OD/SPO sometimes has to be done item-by-item, rather than simply deleting the entire folder. There is an arg `by_item` which can be passed to pin_delete/pin_version_delete, but I'm not sure where to document it. I've added a note in the body for `?pin_delete` and `?pin_version_delete` for this.

Similarly I added a `...` arg to `pin_versions_prune` so that `by_item` can be passed in.

hongooi73:
Tests pass locally. How I've set it up is:
1. Create a drive object eg with `get_personal_onedrive()` or `get_business_onedrive()`
2. Save it somewhere
3. Put the location of the file in the env var PINS_MS365_TEST_DRIVE

This probably isn't the best way to do it in a pipeline, but was easiest to code up. Also, the drive has to allow deleting non-empty folders as noted above. A personal OneDrive should be ok; a SharePoint site or business OneDrive may have doc protection policies.

R/board_ms365.R
```diff
@@ -0,0 +1,183 @@
...
+  try(drive$create_folder(path), silent = TRUE)
+  folder <- drive$get_item(path)
+  if (!folder$is_folder()) {
+    abort("Invalid path specified")
+  }
+
+  cache <- cache %||% board_cache_path(paste0("ms365-", hash(folder$properties$id)))
+  new_board_v1("pins_board_ms365",
+    folder = folder,
+    # path = path,
```
hadley, line 44:
```suggestion
```

R/pin-delete.R
```diff
@@ -4,6 +4,8 @@
 #'
 #' @inheritParams pin_read
 #' @param names The names of one or more pins to delete
+#' @details
+#' Some boards require additional arguments when deleting pins. In particular, deleting pins in OneDrive and SharePoint Online (`board_onedrive`, `board_sharepoint` and `board_ms365`) may require passing `by_item=TRUE` to work around document protection policies that prohibit deleting non-empty folders.
```
hadley, line 8: I think it's probably better to put this in the docs for `board_ms365()`?

R/board_ms365.R
```diff
@@ -0,0 +1,183 @@
...
+  ms365_list_dirs(board)
+}
+
+#' @export
+pin_exists.pins_board_ms365 <- function(board, name, ...) {
+  name %in% ms365_list_dirs(board)
+}
+
+#' @export
+pin_delete.pins_board_ms365 <- function(board, names, by_item = FALSE, ...) {
```
hadley, line 76: Is it possible to catch the error that happens when document protection is on and provide a better error message? (I'd only consider doing this if `ms365_delete_dir()` throws a classed error; string matching on error text is too fragile.)

hadley (commented):
Thanks for working on this! I'll take a more thorough look next week, and figure out how to get the tests working on GHA. 

How do I make a OneDrive? (Or do I already have one with my MS 365 subscription?)

hongooi73:
There are a number of ways to get access to a OneDrive/Sharepoint doc library:

- You can go to https://onedrive.live.com and check if you already have a personal OneDrive. The first 5GB is free for everyone. Once you're setup, you can access the storage via `get_personal_onedrive()`.
- If you have a personal MS 365 sub, that comes with OneDrive. Again, access this with `get_personal_onedrive()`.
- If your org has a commercial MS 365 sub, it will most likely have Sharepoint as part of it. You can see the list of sites you have access to with `list_sharepoint_sites()`. Pick one, then run its `get_drive()` method.
- Sharepoint also enables OneDrive for Business, so if you have SPO, you also have ODB. Access your ODB with `get_business_onedrive()`.

R/pin-delete.R
hongooi73: My thinking was, if I get an error on `pin_delete`, I'll most likely look in `?pin_delete` for help. Maybe put it in both places?

R/board_ms365.R
hongooi73: The error message I've seen isn't too bad; it actually tells you that you can't delete a folder that contains files. But I'm not familiar with all the possible error messages that MS 365 can produce.

The other possibility is to switch the default to `by_item=TRUE`, which should work everywhere. It'll make deleting folders that contain lots of files rather slow, but maybe that's not a problem here.

hongooi73:
A more "industrial-strength" testing solution might be something like the following:

```r
tenant <- Sys.getenv("PINS_MS365_AAD_TENANT")
app <- Sys.getenv("PINS_MS365_CLIENT_ID")
pwd <- Sys.getenv("PINS_MS365_CLIENT_SECRET")
drive_id <- Sys.getenv("PINS_MS365_DRIVE_ID")

# create a Microsoft Graph client
gr <- AzureGraph::create_graph_login(tenant=tenant, app=app, password=pwd, auth_type="client_credentials")

# get the drive directly without going through a user account or Sharepoint site
drv <- gr$get_drive(drive_id)

board_ms365(drv, path = "pin_testing", cache = tempfile(), ...)
```

But that requires some work on the backend to create an app registration, give it the necessary permissions to talk to Microsoft Graph, and get the ID of the drive.

R/board_ms365.R
hongooi73: I've revamped it so that by-item deletion is now an attribute of the board itself. This means you no longer have to pass an argument to `pin_delete` and `pin_version_delete`, which should be a much cleaner solution all-round.

R/pin-delete.R
hongooi73: `delete_by_item` is now an argument to `board_ms365`, so passing an arg to `pin_delete` is no longer necessary.

R/board_ms365.R
hadley: Perfect!

R/board_ms365.R
```diff
@@ -0,0 +1,188 @@
...
+    folder = folder,
+    cache = cache,
+    versioned = versioned,
+    delete_by_item = delete_by_item
+  )
+}
+
+#' @export
+#' @rdname board_ms365
+board_sharepoint <- board_ms365
```
hadley, line 58: I'm a bit conflicted about these: on the one hand they are more explicit names, on the other hand, they do exactly the same thing which might be confusing for users. I think on the whole I'd prefer to stick with just `board_ms365`.

R/board_ms365.R
```diff
@@ -0,0 +1,188 @@
...
+
+#' @export
+#' @rdname board_ms365
+board_onedrive <- board_ms365
+
+board_ms365_test <- function(...) {
+  skip_if_missing_envvars("board_ms365()", "PINS_MS365_TEST_DRIVE")
```
hadley, line 65: How did you make this object? Is there some way I could extract data out of `Microsoft365R::get_personal_onedrive()` to get a re-usable token, or will this always require re-auth?  Your full "industrial solution" seems a bit heavy — is there something that's maybe a bit hacky but is lighter weight?

hadley (commented):
Overall looks good, and my only real concern is figuring out how to test this so I don't accidentally break it in the future. 

Do you want to also mention/list in `_pkgdown.yml`, `NEWS.md` and `README.Rmd`?

I'm going to aim to submit to CRAN on Friday, so let me know if you don't have time between now and then and I'll finish it off.

R/board_ms365.R
hongooi73: It's pretty simple:

```r
od <- get_personal_onedrive()
saveRDS(od, "path/to/file.rds")
Sys.setenv(PINS_MS365_TEST_DRIVE="path/to/file.rds")
```

I assume you're testing S3 and Azure storage in the pipeline? You could copy the rds file into a bucket/container and download it from there as part of the pipeline. It's an R6 object and pretty much self-contained.

If you don't want to copy the full drive object, the OAuth token itself is in the `$token` component. It has a very similar structure to a `httr::Token` object. You can pass this to `get_personal_onedrive` to recreate the drive object:

```r
token <- od$token
saveRDS(token, "path/to/file.rds")

...

token_in_testing <- readRDS("file.rds")
od_in_testing <- get_personal_onedrive(token=token_in_testing)
```

The main caveat is that the embedded refresh token has a [default lifetime](https://docs.microsoft.com/en-us/azure/active-directory/develop/refresh-tokens) of 90 days. This means that as long as you're running the pipeline every 90 days, everything will work; after that, the token will run out and you'll have to reupload a new object. I'm also not entirely sure if it's kosher to be copying tokens from one machine to another, but nothing seems to complain.

hongooi73:
Done. Also simplified deletion a bit more.

Thanks for making this package, and for making it so easy to extend!

hongooi73:
@hadley I just realised I might need to add some logic to properly handle boards in OneDrive. The problem is that OD is only accessible to yourself; while you can share files and folders with others, this has to be handled as a special case.

How do you handle this scenario with Dropbox and Google Drive? Do you just assume such boards will only have one user?

hadley:
@hongooi73 I just assume that the user takes care of that, using the appropriate sharing UI.

R/board_ms365.R
hadley: Ok, I've made a small hack to make interactive testing easier (set an env var to allow using your personal one drive) and a note to look into this again when I have more time (#482)

hongooi73:
So, a couple of things:

First, I realised that the simple testing setup above, where we copy a token/drive object around, will only work for 90 days max. After that, the token will expire and since we never save a new refresh token, the expiry is permanent. I don't think there's any easy way to avoid the industrial-strength solution. 🤔

2nd, I've found a way to share a board in OneDrive, but...
- It's rather messy, requiring the recipient to locate the board from the list of shared item objects, then pass the object as the `path` argument. See the rewritten docs.
- It'll only work with the next version of Microsoft365R, as the current version has a bug 😅

Thoughts? I was thinking that I should leave the docs out but keep the functionality in, since sharing in OneDrive won't actually work right now. Note that Sharepoint is fine, since all users should have access to the board.

hadley:
>  I was thinking that I should leave the docs out but keep the functionality in, since sharing in OneDrive won't actually work right now. 

Yeah that sounds good to me.

hongooi73:
Actually, it turns out that a shared OneDrive folder WILL work after all, I just need to sync the data up.

The PR should be ready to go. The interface is a bit inelegant; if you supply a drive item (R6 object) as the `path`, the `drive` arg is not used. In other languages that support function overloading, you could have 2 overloads: `board_ms365(drive, path)` and `board_ms365(drive_item)`, but I'm not sure about the best way to do it here.

hadley:
Thanks for all your work on this! I'm still on track for a CRAN release tomorrow.

PR #514, opened 2021-09-13
Add qs format

smingerson:
There are arguments in both [`qsave()`](https://rdrr.io/cran/qs/man/qsave.html) and [`qread()`](https://rdrr.io/cran/qs/man/qread.html) which can affect the performance of reading and writing. Regardless, [qs performs better than RDS](https://github.com/traversc/qs), so I don't think providing a mechanism to change the defaults is needed, at least not initially.

### DIFF
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -46,7 +46,7 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   When retrieving the pin, this will be stored in the `user` key, to
 #'   avoid potential clashes with the metadata that pins itself uses.
 #' @param type File type used to save `x` to disk. Must be one of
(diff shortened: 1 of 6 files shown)


R/pin-read-write.R
```diff
@@ -111,14 +111,15 @@ guess_type <- function(x) {
 }
 
 object_write <- function(x, path, type = "rds") {
-  type <- arg_match0(type, c("rds", "json", "arrow", "pickle", "csv"))
+  type <- arg_match0(type, c("rds", "json", "arrow", "pickle", "csv", "qs"))
```
hadley, line 114: Also need to update docs above; and might make sense to change this to `setdiff(object_types, "file")`

hadley (commented):
Looking good, thanks!

hadley:
Thanks @smingerson!

PR #516, opened 2021-09-14
disable rsc workflow if the license is empty

colearendt:
Having a hard time determining that "this PR is coming from a fork"

- `github.repository` is always set to `rstudio/pins`
- `secrets` is not available in the `if` block


hadley:
@colearendt I'm going to try this approach that @maelle figured out for pkgdown

maelle:
For the record I'm fairly positive I copied an approach of @jennybc's e.g. in https://github.com/hadley/r-pkgs/blob/master/.github/workflows/build-book.yaml :grin:

PR #520, opened 2021-09-22
Update RSC test boards

hadley:
Since they never actually used envvars, and this now (correctly) fails. And improve auth="envvar" option to give better errors


juliasilge (approved):
That is the error I get in that situation ✅

PR #522, opened 2021-09-23
Normalize path

### LINKED ISSUE #510: Errors using board_folder() with VSCode+Python+radian
I'm not sure if this is purely a pins bug, or an unexpected interaction with my IDE setup. I'm trying to create a simple local board:

```r
f <- "~/misc/pins/tempboard3"
dir.create(f, recursive=TRUE)
b <- board_folder(f)
b %>% pin_write(mtcars, "mtcars")
```
This results in
```
Guessing `type = 'rds'`
Creating new version '20210912T095121Z-0fc2e'
Error in file(file, "w", encoding = fileEncoding) : 
  cannot open the connection
In addition: Warning message:
In file(file, "w", encoding = fileEncoding) :
  cannot open file 'C:\Users\hongo\Documents/misc/pins/tempboard3/mtcars/20210912T095121Z-0fc2e/data.txt': No such file or directory
```
Indeed, I can see that the 'tempboard3' folder (I previously tried tempboard and tempboard2, changing the names each time to ensure I wasn't picking up a stale cache) is empty. However, there is apparently something in `b`:

```
Pin board <pins_board_folder> 
Path: '~/misc/pins/tempboard3'
Cache size: 0
Pins [1]: 'mtcars'
```
I went looking for the cache folder, and eventually found it in: C:\Users\hongo\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\Local\pins

which is a location owned by the Python app from the Microsoft store. Python is involved because I use [radian](https://github.com/randy3k/radian), which is a Python-powered R interpreter. I'm not sure why it's there, I'd have expected the cache to sit under an R-specific location. In any case, deleting that folder and trying again still gave the above error.

hadley:
Fixes #510

@hongooi73 can you please see if this helps?

### DIFF
--- a/R/board_folder.R
+++ b/R/board_folder.R
@@ -23,6 +23,7 @@
 #' @export
 board_folder <- function(path, versioned = FALSE) {
   fs::dir_create(path)
+  path <- fs::path_norm(path)
 
   new_board_v1("pins_board_folder",
     cache = NA_character_,


hongooi73:
Yup, works fine!

PR #523, opened 2021-09-24
Add `board_deparse` generic

juliasilge:
This PR adds a `board_deparse` generic that will generate a character string of R code from a board object, like so:

``` r
library(pins)

b <- board_rsconnect()
#> Connecting to RSC 1.9.0.1 at <https://colorado.rstudio.com/rsc>
board_deparse(b)
#> board_rsconnect(server = "colorado.rstudio.com")
```

<sup>Created on 2021-09-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This will be helpful for when we use a pin locally and need to write code to access the same pin from another machine.

This draft right only implements `board_deparse` for RStudio Connect and shows how we would use it in the RSC bundle. I put the generic in `board.R` since it applies to all boards but I could make a new `deparse.R` if you prefer.

Still TODO, after any changes needed here:

- Rest of boards
- Add tests

Anything else?


R/board_rsconnect.R
```diff
@@ -324,6 +324,12 @@ pin_search.pins_board_rsconnect <- function(board, search = NULL, ...) {
   multi_meta(board, paste0(user, "/", name))
 }
 
+
+#' @export
+board_deparse.pins_board_rsconnect <- function(board, ...) {
+  glue('board_rsconnect(server = "{board$server}")')
```
juliasilge, line 330: Honestly a little unsure if this should be `server` or `server_name`

hadley: We'll probably need to think this through carefully. I suspect it depends on exactly which authentication you used 😬

inst/preview/index.html
```diff
@@ -28,7 +28,7 @@
     <section>
     <h3>Code</h3>
       <pre id="pin-r" class="pin-code"><code class="r">library(pins)
-board <- board_rsconnect(server = "{{server_name}}")
+board <- {{board_deparse}})
```
hadley, line 31:
```suggestion
board <- {{{board_deparse}}})
```

hadley (commented):
Looks like a good start and 100% in line with what I was thinking 😄

R/board_rsconnect.R
juliasilge: OMG `board$server` only works in my particular situation because of partial matching 💀 

```r
> names(b)
 [1] "board"                "api"                  "cache"                "name"                
 [5] "url"                  "account"              "server_name"          "auth"                
 [9] "versioned"            "use_cache_on_failure"
```

juliasilge: Take a look at how I approached this; I think it will work for the various auth strategies there are for RSC.

juliasilge:
I have this working pretty nicely for S3, I think:

``` r
library(pins)

b <- board_s3("pins-test-julia", region = "us-east-2")
board_deparse(b)
#> board_s3(bucket = "pins-test-julia", region = "us-east-2")
```

<sup>Created on 2021-09-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


I don't have the env variables and such to be able to regenerate the snapshots for RSC and S3. Why is the CI all green if I added tests but didn't update the snapshots? (I did update snapshots for `board_local()` and `board_url()` which by the way just errors.)



Still TODO is Azure, for which I don't have an account so I'll be doing a bit of blindly stabbing in the dark.

Is it useful to have this for Kaggle?

R/board_s3.R
```diff
@@ -221,6 +221,34 @@ pin_store.pins_board_s3 <- function(board, name, paths, metadata,
...
+  }
+
+  ## non-auth config
+  config <- board$svc$.internal$config
+  region <- config$region
+  if (nchar(region) == 0) region <- NULL
+  endpoint <- config$endpoint
+  if (nchar(endpoint) == 0) endpoint <- NULL
+  profile <- config$credentials$profile
+  if (nchar(profile) == 0) profile <- NULL
```
juliasilge, line 241: I don't have quite enough experience with AWS to know if `profile` is more like server-level config or more like auth. Should it stay with the board or should we expect people to have it on their remote server in something like an environment variable?

juliasilge:
Here are some open questions to answer/confirm/address:

- `board_deparse()` is not implemented for Kaggle for now (can come back to it if users would find it useful)
- The snapshots need to be regenerated (I don't have the right keys, etc)
- Should `profile` be included on AWS or not?

hadley:
* For `board_kaggle_*()`, I think you can just assume that `username`/`key` will come from a `kaggle.json` so the method can be very simple
* Will regenerate momentarily. CI looks good because generating the first snapshot is only a warning, not a failure. I wonder if it [should be an error on CI](https://github.com/r-lib/testthat/issues/1461)?
* I'll check with Joe, but think `profile` should not be included — I suspect a common use would be `profile = "production"` vs `profile = "development"`

R/board_s3.R
```diff
@@ -221,6 +221,34 @@ pin_store.pins_board_s3 <- function(board, name, paths, metadata,
...
+  if ("prefix" %in% names(board)) {
+    prefix <- board[["prefix"]]
+  } else {
+    prefix <- NULL
+  }
+
+  ## non-auth config
+  config <- board$svc$.internal$config
+  region <- config$region
+  if (nchar(region) == 0) region <- NULL
```
hadley, line 237:
```suggestion
  region <- if (nchar(region) == 0) NULL else region
```

I think this form is fine for simple expressions

R/board_s3.R
```diff
@@ -221,6 +221,34 @@ pin_store.pins_board_s3 <- function(board, name, paths, metadata,
...
+  endpoint <- config$endpoint
+  if (nchar(endpoint) == 0) endpoint <- NULL
+  profile <- config$credentials$profile
+  if (nchar(profile) == 0) profile <- NULL
+
+  board_args <- c(bucket = bucket, prefix = prefix, region = region,
+                  endpoint = endpoint, profile = profile)
```
hadley, line 244: Can you please style as in https://style.tidyverse.org/syntax.html#long-lines ?

R/board_url.R
```diff
@@ -139,6 +139,12 @@ pin_store.pins_board_url <- function(board, name, paths, metadata,
   abort_board_read_only("board_url")
 }
 
+#' @rdname board_deparse
+#' @export
+board_deparse.pins_board_url <- function(board, ...) {
```
hadley, line 144: There's no reason this couldn't be deparsed, although I don't think you need to implement it, so I think you can just rely on the default behaviour here.

R/board_folder.R
```diff
@@ -117,3 +117,10 @@ pin_versions.pins_board_folder <- function(board, name, ...) {
 pin_version_delete.pins_board_folder <- function(board, name, version, ...) {
   fs::dir_delete(fs::path(board$path, name, version))
 }
+
+#' @rdname board_deparse
+#' @export
+board_deparse.pins_board_folder <- function(board, ...) {
+  path <- check_board_deparse(board, "path")
+  glue('board_folder(path = "{path}")')
```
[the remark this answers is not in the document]
juliasilge: I could see using `expr_deparse()` for some of these, but I couldn't see a straightforward way to make it work for a complex one like S3. Want to take a look and see if I am missing it? Would you want some `glue()` and some `expr()`?

R/board_azure.R
```diff
@@ -137,6 +137,18 @@ pin_store.pins_board_azure <- function(board, name, paths, metadata,
...
 }
 
+
+#' @rdname board_deparse
+#' @export
+board_deparse.pins_board_azure <- function(board, ...) {
+  container <- check_board_deparse(board, "container")
+  blob_url <- container$endpoint$url
+  blob_name <- container$name
+  container <- glue('AzureStor::blob_container(endpoint = AzureStor::storage_endpoint("{blob_url}"), name = "{blob_name}")')
```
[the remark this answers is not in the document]
hongooi73: This is kind of a weird request, lol.

I think the best solution is to generate

```
board_azure(AzureStor::storage_container("url", key=NULL, token=NULL, sas=NULL), "path/to/board")
```
where the `key`, `token` and `sas` args are left blank. The other user then has to supply at least one of these to authenticate.

The `key` is the storage account primary key, and you **really** don't want to be passing that around. (You shouldn't be using it at all, but AzureStor lets you if you really want to.)

`token` is an OAuth token object from AzureAuth, and also shouldn't be shared. If you authenticated with an interactive flow, it identifies you as the user, and giving it to someone else violates basic principles of authentication.

You could probably pass `sas` around, but it feels rather dodgy to me. If the other user is authorised to use the storage account, they should already have a SAS of their own, so they can use that.

The URL of the storage container can be found from the `endpoint$url` and `name` components:

```
r$> cont <- storage_container("https://myacct.blob.core.windows.net/cont", key="key")   
r$> cont
Azure blob container 'cont'
URL: https://myacct.blob.core.windows.net/cont
Access key: <hidden>
Azure Active Directory access token: <none supplied>
Account shared access signature: <none supplied>
Storage API version: 2020-04-08
r$> cont$endpoint$url
[1] "https://myacct.blob.core.windows.net/"
r$> cont$name
[1] "cont"
```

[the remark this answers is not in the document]
hongooi73: Also, I think I'll have to add some more bits to `board_ms365` to support deparsing like this. The object returned by the Graph API mungs things up something fierce.

R/board_folder.R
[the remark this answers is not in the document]
hadley: I'm trying out `expr()` and while it doesn't do great with the wrapping, it feels like a nice win overall. Changes incoming soon!

R/board_rsconnect.R
```diff
@@ -324,6 +324,20 @@ pin_search.pins_board_rsconnect <- function(board, search = NULL, ...) {
   multi_meta(board, paste0(user, "/", name))
 }
 
+
+#' @rdname board_deparse
+#' @export
+board_deparse.pins_board_rsconnect <- function(board, ...) {
+  if (has_name(board, "url")) {
+    server <- board$url
```
hadley, line 332: When will `url` not be present? 

I think this also needs a fix in `rsc_server()` which I'll make momentarily.

R/board_azure.R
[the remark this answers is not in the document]
hadley: Thanks!

hadley:
@juliasilge can you take one last pass through this please? (I'm aiming for submission on Friday.

R/board.R
```diff
@@ -132,6 +132,34 @@ board_cache_path <- function(name) {
   fs::path(path, name)
 }
 
+#' Deparse a board object
+#'
+#' Returns the R code that would recreate the board is re-run on another
+#' computer. Goal is to capture the location of the board, but not the
+#' authorisation, since (a) that leak would leak credentials and (b) in
+#' most deployment scenarios board auth will be read from env vars.
```
juliasilge, line 140:
```suggestion
#' Returns the R code that would recreate the board when re-run on another
#' computer. Goal is to capture the location of the board, but not the
#' authorisation, since (a) that would leak credentials and (b) in
#' most deployment scenarios board auth will be read from env vars.
```
I think??

R/board_rsconnect.R
juliasilge: I thought I got into that situation when I used `auth = "rsconnect"` that had `server_name` but not `url` but now I am having trouble reproducing. Can you confirm that `url` is always there and sufficient?

juliasilge:
I installed from this branch again and ran through all the setups I have -- everything is working 👍

R/board_rsconnect.R
hadley: Yeah, I'm pretty sure — `url` comes from `server$url` which comes from `rsc_server()` which calls either `rsc_server_manual()` or `rsc_server_rsconnect()`, and both set the `url`.

Issue #528, opened 2021-10-06
`board_azure` should allow setting the path to the board


hongooi73:
Currently, `board_azure` doesn't have any facility to specify where the board is located inside the container. However, all storage services that AzureStor supports (file, blob, ADLS2) can have directories, like with a local disk, so there is no reason not to allow it.

I can make a PR for this.

hongooi73:
This is basically the equivalent of specifying a prefix with S3; you can actually do the same with Azure, but using directory syntax seems more straightforward and also generalises to file and ADLS2 storage.

hadley:
Yeah, a PR would be great.

PR #532, opened 2021-10-13
Enhancements to `board_azure`

hongooi73:
This PR makes the following changes:

- Allow creating boards in storage services other than blob storage (file and ADLS2). The recommendation is to use ADLS2 over blob storage if possible, as the backend is much better at handling directories
- Allow creating boards in locations other than the root directory, using the `path` arg to specify a subdirectory
- Added tests for all storage services, as well as creating boards in both the root and non-root locations. The new tests are in the files `test-board_azure_adls2.R`, `test-board_azure_blob.R` and `test-board_azure_file.R`.

Fully implementing the first point was rather messy due to underlying differences in API behaviour, as well as some missing functionality in AzureStor (recursive deleting of non-empty subdirs), but it should all be working now. I'll be backporting the code to AzureStor so in the future we should be able to clean the pins code a bit.

I also changed the `board_azure_test` function to allow specifying the storage account name in the PINS_AZURE_ACCOUNT env var, so that I could run the tests on my own account. If this var is unset, it defaults to "pins" which is the account currently used for testing.

I haven't added new snapshots for the tests, since I'm using my own storage account. However, the outputs should look like ` board_azure(AzureStor::storage_container("https://account_url"), path="test/path")`


hongooi73:
closes #528

R/board_azure.R
```diff
@@ -1,12 +1,18 @@
...
 #'
 #' @inheritParams new_board
 #' @param container An azure storage container created by
 #'   [AzureStor::blob_container()] or similar.
+#' @param path Path to directory to store pins. Will be created if it
+#'   doesn't already exist.
 #' @param n_processes Maximum number of processes used for parallel
 #'   uploads/downloads.
+#' @details
+#' You can create a board in any of the services that AzureStor supports: blob storage, file storage and Azure Data Lake Storage Gen2 (ADLSgen2).
```
hadley, line 13: Can you wrap these to 80 chars please?

R/board_azure.R
```diff
@@ -20,31 +26,59 @@
...
+#' # - Use 'dfs' instead of 'blob' in the account URL to use the ADLSgen2 API
+#' # - We reuse the board created via the blob API above
+#' adls_url <- "https://myaccount.dfs.core.windows.net/mycontainer"
+#' container <- AzureStor::storage_container(adls_url, sas = "my-sas")
+#' board <- board_azure(container, "path/to/board")
+#' board %>% pin_list()
+#' board %>% pin_read("iris")
 #' }
-board_azure <- function(container, n_processes = 10, versioned = TRUE, cache = NULL) {
+board_azure <- function(container, path = "", n_processes = 10, versioned = TRUE, cache = NULL) {
```
hadley, line 44: I think you need to document `path` since it's inheriting a description that doesn't make sense here.

And maybe it should be called `prefix` to match `board_s3()`?

R/board_azure.R
```diff
@@ -20,31 +26,59 @@
...
 
-  cache <- cache %||% board_cache_path(paste0("azure-", hash(container$endpoint$url)))
+  if (path == "/") {
+    path <- ""
+  }
+  board_path <- fs::path(container$endpoint$url, container$name, path)
+  cache <- cache %||% board_cache_path(paste0("azure-", hash(board_path)))
+  if (path != "") {
+    if(inherits(container, "file_share")) {
+      try(AzureStor::create_storage_dir(container, path, recursive = TRUE), silent = TRUE)
```
hadley, line 54: Why the try?

R/board_azure.R
```diff
@@ -145,38 +179,83 @@ board_deparse.pins_board_azure <- function(board, ...) {
...
 
 
 # Helpers -----------------------------------------------------------------
 
 azure_delete_dir <- function(board, dir) {
-  ls <- AzureStor::list_storage_files(board$container, dir)
-  for (path in ls$name) {
-    AzureStor::delete_storage_file(board$container, path, confirm = FALSE)
+  dir <- azure_normalize_path(board, dir)
+
```
hadley, line 190: Maybe add comment with link to corresponding AzureStor issue if it exists? (Just to remind us to eventually simplify this code once the logic is in AzureStor)

R/board_azure.R
```diff
@@ -20,31 +26,59 @@
...
   check_installed("AzureStor")
 
-  cache <- cache %||% board_cache_path(paste0("azure-", hash(container$endpoint$url)))
+  if (path == "/") {
+    path <- ""
+  }
+  board_path <- fs::path(container$endpoint$url, container$name, path)
+  cache <- cache %||% board_cache_path(paste0("azure-", hash(board_path)))
+  if (path != "") {
+    if(inherits(container, "file_share")) {
```
hadley, line 53:
```suggestion
    if (inherits(container, "file_share")) {
```

R/board_azure.R
```diff
@@ -145,38 +179,83 @@ board_deparse.pins_board_azure <- function(board, ...) {
...
+  } else if (inherits(board$container, "file_share")) {
+    ls <- AzureStor::list_storage_files(board$container, dir, recursive = TRUE)
+    for (i in rev(seq_len(nrow(ls)))) {
+      if(ls$isdir[i]) {
+        AzureStor::delete_storage_dir(board$container, ls$name[i], confirm = FALSE)
+      } else {
+        AzureStor::delete_storage_file(board$container, ls$name[i], confirm = FALSE)
+      }
+    }
+    if(dir != "/") {
```
hadley, line 212:
```suggestion
    if (dir != "/") {
```

R/board_azure.R
```diff
@@ -197,3 +277,20 @@ azure_download <- function(board, keys, progress = !is_testing()) {
 
   invisible()
 }
+
+azure_upload_file <- function(board, src, dest) {
+  if(inherits(board$container, "file_share")) {
```
hadley, line 282:
```suggestion
  if (inherits(board$container, "file_share")) {
```

tests/testthat/test-board_azure_adls2.R
```diff
@@ -0,0 +1,14 @@
+board_adls <- board_azure_test(path = "", type = "dfs")
```
hadley, line 1: Can you please put all the tests together in `test-board_azure.R`? Our convention is to match the test file name to the `R/` file name

hadley (commented):
Thanks for working on this! A few minor comments/questions below.

R/board_azure.R
hongooi73: This is documented up in line 8. It's `path` by analogy to the path for a local board. "Prefix" is an S3-ism which was copied by blob storage, but it doesn't really apply for the other storage services. I'll clarify this.

R/board_azure.R
hongooi73: The different storage services all behave in annoyingly different ways. For file storage, creating a dir that already exists is an error; the other services will silently ignore it.

I could probably do a better job of providing a consistency layer in AzureStor, but I don't think there's many people who want to do the exact same operations on all 3 services, like here.

R/board_azure.R
hongooi73: Not sure which bit you're referring to here, can you clarify?

tests/testthat/test-board_azure_adls2.R
hongooi73: I put them in different files because it was actually easier this way: I can run tests on one service, without having to repeat all the tests for all the other services. Since the tests can take minutes to run, this is way faster.

I would suggest breaking your convention just this once. 😉

hadley: If they take minutes to run the tests, then they probably need to be handled differently.

hongooi73: Ok, I misremembered. I ran the tests again, and each file (azure_adls2, azure_blob, azure_file) takes about 30 seconds to run.

I'm actually running your standard test suite twice for each service: once for a board in the root directory, and again for a board in a supplied path. Testing both scenarios has actually been useful, there are some corner cases in how each service handles paths that I had to nail down.

Waiting 30 seconds for the test to conclude is better than waiting 1.5 minutes, so I still think it's worthwhile to separate them out. Unless there's a way to run only a subset of the tests in a file?

hongooi73: Oh, one more thing: the SAS generated by the Azure portal won't work for the ADLS2 tests because it uses an outdated API version. You can use the AzureStor package to generate it instead.

```r
library(AzureStor)
endp <- storage_endpoint("https://xxxx.dfs.core.windows.net", key="yyyy")
sas <- get_account_sas(endp, permissions="rwcdl")
Sys.setenv(PINS_AZURE_SAS=sas)
```

Should I add a comment to this effect somewhere?

R/board_azure.R
hadley: Ok, I missed that. I think the explanation and variable name makes sense.

R/board_azure.R
hadley: I thought in your other comment you were implying that this complicated delete code would eventually live in AzureStore?

tests/testthat/test-board_azure_adls2.R
hadley: Ok, leave them as is for now; I think I'll probably set up something custom to skip these tests for me when I'm developing locally and not touching this specific code (i.e. something with `skip_if` and an env var)

R/board_azure.R
hongooi73: Oh right. Yeah, the delete code is actually implemented in AzureStor now, as of a couple months ago. I'll add a comment to this effect.

tests/testthat/test-board_azure_adls2.R
hongooi73: Right now the tests should be skipped if all the necessary env vars aren't set. The way I usually do it is to have a batch script that sets these vars; if I want to skip tests that require auth, I just don't run the script beforehand. Your workflow may be different though.

hadley:
Thanks!

PR #555, opened 2021-11-05
Updates Connect preview page with improved metadata formatting

### LINKED ISSUE #552: yaml meta data displayed as json on the preview page
the metadata in `data.txt` is yaml format, but the metadata is displayed as json on the preview page.

This is confusing, since the data is not readily available in the display format.

The display format should match the storage format.

### LINKED ISSUE #554: Preview page metadata display improvements
The preview page doesn't currently do much with the metadata.

It would be great if the name/title and description could be displayed more prominently.

For instance, at the top of the page (above the code block) we could have something like this:

---

<h2>Palmer Penguins: a pinned 344 x 8 data frame</h2>
Last updated: 2021-08-25.       Format: CSV.        API: 1.0
<h4>This is the description of my pin, it's a very nice pin with lots of penguin data in it. I think the Fiordland penguin is my favourite.</h4>

---

The code block could then live under that.

We could have the raw yaml metadata hidden with a `<details>` as well.

sellorm:
Prioritises the display of important metadata like the "created" timestamp and the description.

Raw metadata is now displayed in its original yaml format but hidden from the default view behind a `<details>` tag.

Closes #554
Closes #552

### DIFF
--- a/R/board_rsconnect_bundle.R
+++ b/R/board_rsconnect_bundle.R
@@ -59,7 +59,11 @@ rsc_bundle_preview_index <- function(board, name, x, metadata) {
     data_preview = jsonlite::toJSON(data_preview, auto_unbox = TRUE),
     data_preview_style = if (is.data.frame(x)) "" else "display:none",
     pin_name = paste0(board$account, "/", name),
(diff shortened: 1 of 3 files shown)


sellorm:
Re-submitted in #558

PR #558, opened 2021-11-11
Improves RSConnect html preview

### LINKED ISSUE #552: yaml meta data displayed as json on the preview page

### LINKED ISSUE #554: Preview page metadata display improvements

sellorm:
Prioritises the display of important metadata like the "created" timestamp and the description.

Raw metadata is now displayed in its original yaml format but hidden from the default view behind a `<details>` tag.

Closes #554
Closes #552


inst/preview/index.html
```diff
@@ -25,6 +25,22 @@
...
   </head>
   <body>
+
+    <section>
+       <h3>{{{pin_name}}}</h3>
+       <p>
+         Last updated: {{{pin_metadata_date}}} •
+         Format: {{{pin_metadata_format}}} •
+         API: v{{{pin_metadata_api_version}}}
+       </p>
```
hadley, line 35: You can use sections (see https://mustache.github.io/mustache.5.html) to avoid having to duplicate the contents of `pin_metadata` (except maybe for the date which you might still need to format)

tests/testthat/_snaps/board_rsconnect_bundle.md
```diff
@@ -34,6 +34,23 @@
           </style>
         </head>
         <body>
+      
+          <section>
+             <h3>TEST/test</h3>
+             <p>
+               Last updated:  •
```
hadley, line 41: Would be good to tweak the test so that we get semi-realistic data here

hadley (commented):
Thanks for working on this!

inst/preview/index.html
sellorm: I'm not sure that will work in this case as `pin_metadata` is stored `as.yaml` so that it can be properly rendered in the `<pre></pre>` section of the template.

I could implement moustache sections if `pin_metadata` was a real R list but still have a duplicate using `as.yaml` in a `pin_metadata_yaml` if that would make things clearer?

tests/testthat/_snaps/board_rsconnect_bundle.md
sellorm: Agreed - I've updated the test accordingly

inst/preview/index.html
hadley: Oh yeah, of course. Maybe something like this:

```R
metadata = list(
  as_yaml = yaml::as.yaml(metadata),
  date = parse_8601_compact(metadata$created),
  format = metadata$type,
  api_version = metadata$api_version,
  description = metadata$description
)
```

sellorm: Done - but I might have gone a bit over the top, let me know what you think.

This new version also properly improves handling of pre-1.0.0 pins. The date is only displayed if it exists (which it did not prior to 1.0.0) and in the absence of the `api_version` in the metadata 'v0' is displayed instead.

I've also tried to improve readability by making the title bold and adding some padding to the `<pre>` sections.

PR #561, opened 2021-11-15
Make error more helpful if rsconnect package not installed

### LINKED ISSUE #560: Unhelpful error when `rsconnect` not installed
I just ran the following:

```r
> pins::board_rsconnect()
Error: auth = `auto` has failed to find a way to authenticate
* `server` and `key` not provided for `auth = 'manual'`
* Can't find CONNECT_SERVER and CONNECT_API_KEY envvars for `auth = 'envvar'`
* Can't find any rsconnect::accounts() for `auth = 'rsconnect'`
Run `rlang::last_error()` to see where the error occurred.
```

However, I actually do have connect accounts initialized. The problem is that I was in a fresh `renv` project and didn't have `rsconnect` installed. It would be really helpful if this checked if `rsconnect` was installed and suggested installing if it's not.

akgold:
Closes #560

### DIFF
--- a/R/board_rsconnect_server.R
+++ b/R/board_rsconnect_server.R
@@ -22,6 +22,13 @@ check_auth <- function(auth = c("auto", "manual", "envvar", "rsconnect"), server
       "envvar"
     } else if (rsc_rsconnect_is_configured()) {
       "rsconnect"
+    } else if (!is_installed("rsconnect")) {
+      abort(c(
+        "auth = `auto` has failed to find a way to authenticate",
+        "`server` and `key` not provided for `auth = 'manual'`",
+        "Can't find CONNECT_SERVER and CONNECT_API_KEY envvars for `auth = 'envvar'`",
+        "rsconnect package not installed for `auth = 'rsconnect'`"
+      ))
     } else {
       abort(c(
         "auth = `auto` has failed to find a way to authenticate",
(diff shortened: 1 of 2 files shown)


R/board_rsconnect_server.R
```diff
@@ -22,6 +22,13 @@ check_auth <- function(auth = c("auto", "manual", "envvar", "rsconnect"), server
       "envvar"
     } else if (rsc_rsconnect_is_configured()) {
       "rsconnect"
+    } else if (!is_installed("rsconnect")) {
+      abort(c(
```
hadley, line 26: Could you please reduce the duplication with the text in the next block? I'd suggest something like:

```R

if (is_installed("rsconnect")) {
  auth_rsc <- "Can't find any rsconnect::accounts() for `auth = 'rsconnect'"
} else {
  auth_rsc <- "rsconnect package not installed for `auth = 'rsconnect'`"
}


abort(c(
   "auth = `auto` has failed to find a way to authenticate:",
    "`server` and `key` not provided for `auth = 'manual'`",
  "Can't find CONNECT_SERVER and CONNECT_API_KEY envvars for `auth = 'envvar'`",
  auth_rsc,
))
```

akgold:
Updated to be DRY-er. Wasn't sure if that was preferred vs more explicit but wordier :)

Issue #566, opened 2021-11-16
board_ms365() tests are really slow


hadley:
They're taking > 80s for me. @hongooi73 did anything change lately? I don't remember them being this slow when I last worked on pins.

hongooi73:
I haven't changed anything, but Microsoft365R isn't that optimised. The server could also just be slow right now.

Does profiling show any obvious bottlenecks?

lemuelemos:
i'm having the same problem. When a try to pin some large DT i'm having the error: 

"Error in curl::curl_fetch_memory(url, handle = handle) : 
  OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104"

hadley:
@lemuelemos that's not the same problem. Please create a reprex and file a new issue.

hadley:
Hmmm, maybe we should switch to smaller data files? I'm going to be working on pins again this week, so I'll hopefully have refamiliarised myself with what's going on shortly.

hadley:
Looks like there's nothing obvious we can do to improve performance so I'm going to close this. If I get too annoyed by how long it takes these tests to run, I might add some skip functionality based on an env var.

PR #571, opened 2021-12-07
Fix bug in ms365 testing

hongooi73:
This fixes an error that occurs when the testing folder is still lying around from a previous test.

see comment in #566


hadley:
Thanks! (Failures seem to be some temporary failure with Github)

PR #583, opened 2022-01-06
pin_read instead of pin_get (legacy)

duju211:


juliasilge:
Thanks for this PR @duju211! When you contribute [fixes to documentation](https://www.tidyverse.org/contribute/#contribute-documentation), you'll want to make the [change in the `.R` file, not the `.Rd` file](https://tidyverse.tidyverse.org/CONTRIBUTING.html) that is automatically generated via roxygen2.

hadley (approved):

PR #595, opened 2022-02-01
Replacing pin_created with pin_hash

### LINKED ISSUE #542: pins_reactive_read does not update when pin changes more frequently than once a minute
the `pin_reactive_read` function sets up a `shiny::reactivePoll` object using `pin_created` as the `checkFunc`.  However, this is not ideal, because `pin_created` only seems to return a timestamp with accuracy to the minute, and so if a pin is updated twice within the same minute, the reactive is never triggered.  Wouldn't it make more sense to use the hash value?  Here is an example shiny app. 

``` r
pin_hash <- function(name, board) pins::pin_meta(board, name)$pin_hash

custom_pin_reactive_read <- function(board, name, interval) {
  rlang::check_installed("shiny")
  shiny::reactivePoll(intervalMillis = interval, 
                      session = shiny::getDefaultReactiveDomain(), 
                      checkFunc = function() pin_hash(name = name, board = board), 
                      valueFunc = function() pins::pin_read(board = board, name = name))
}

write_time_pin <- function(board) pins::pin_write(board, Sys.time(), "time")

ui <- shiny::fluidPage(shiny::textOutput("pins"),
                       shiny::textOutput("custom"))

server <- function(input, output, session) {
  board <- pins::board_temp()
  write_time_pin(board)
  # update pin once every second
  observe({ 
    write_time_pin(board)
    shiny::invalidateLater(1000)
  })
  
  pins_pin_reactive_read_value <- pins::pin_reactive_read(board = board, "time", 1000)
  custom_pin_reactive_read_value <- custom_pin_reactive_read(board = board, "time", 1000)
  output$pins <- shiny::renderPrint(req(pins_pin_reactive_read_value()))
  output$custom <- shiny::renderPrint(req(custom_pin_reactive_read_value()))
}

shiny::shinyApp(ui, server)
```

thomaszwagerman:
Replaces `pin_created()` with `pin_hash()`, which supports functionality of `pin_reactive_read()` and `pin_reactive_download()`.

Suggestion made by @TylerGrantSmith.

Closes #542

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
   
 * `pin_upload()` now better handles path expansion (#585, @sellorm).
 
+* The `pin_reactive_*()` functions now use the hash (rather than the created 
+  date) for polling (#595, @thomaszwagerman).
+
 # pins 1.0.1
 
 * `board_azure()` now allows you to set a `path` so that multiple boards can
--- a/R/pin_reactive.R
+++ b/R/pin_reactive.R
@@ -72,7 +72,7 @@ pin_reactive_read <- function(board, name, interval = 5000) {
   shiny::reactivePoll(
     intervalMillis = interval,
     session = shiny::getDefaultReactiveDomain(),
-    checkFunc = function() pin_created(board, name),
+    checkFunc = function() pin_hash(board, name),
     valueFunc = function() pin_read(board, name)
   )
 }
@@ -85,12 +85,12 @@ pin_reactive_download <- function(board, name, interval = 5000) {
   shiny::reactivePoll(
     intervalMillis = interval,
     session = shiny::getDefaultReactiveDomain(),
-    checkFunc = function() pin_created(board, name),
+    checkFunc = function() pin_hash(board, name),
     valueFunc = function() pin_download(board, name)
   )
 }
 
-pin_created <- function(board, name) {
+pin_hash <- function(board, name){
   meta <- pin_meta(board, name)
-  meta$created
+  meta$pin_hash
 }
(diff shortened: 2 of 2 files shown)


thomaszwagerman:
Looking at this again and had a thought.

I've replaced `pin_created()` with a function called `pin_hash()`, however where `pin_meta()` returns `$created` for the date the pin was created, for the hash of the pin contents it returns `$pin_hash`. When I looked back over the code this caused confusion.

Either I change the `pin_hash()` function name, or the output of `pin_meta()` should be changed from `$pin_hash` -> `$hash`, which I think is more consistent with the rest of `pin_meta()`'s outputs.

juliasilge:
Thank you so much for this PR @thomaszwagerman and for your patience on it!

I agree that it's not ideal that the hash is called `pin_hash` in the pin metadata. However, it wouldn't really solve this specific question because there is an existing unexported function called `pin_hash()`:

https://github.com/rstudio/pins-r/blob/9d014de4d9384213ebc4e6e02888de5eb36d4ffd/R/pin-read-write.R#L209-L216

Even if we were to change the name of `pin_hash`, we still need to change the function name here. Maybe `get_pin_hash()`?

thomaszwagerman:
Hi @juliasilge, thank you for you reply, and no worries at all for the wait!

Apologies, I was not aware of the unexported `pin_hash()` - I've implemented your suggestion by changing the function name to `get_pin_hash()`.

juliasilge (approved):
Thank you so much for this PR @thomaszwagerman! 🙌

PR #618, opened 2022-05-23
Change repo to "pins-r"

juliasilge:
This PR sets up changes needed to change the repo name to "pins-r", for parity with https://github.com/rstudio/pins-python. I tried hard to find everything that would need to be changed, but another set of eyes on this will be great. Also, the changes here will not pass CI until the repo is actually renamed.


juliasilge:
@machow could you take a look at this and see if you find any other places I need to change the GitHub repo name?

machow:
I think since we were able to get tests passing in a PR (https://github.com/rstudio/pins/pull/619), and because the failing checks are due to trying to pull data from the new repo name, it seems okay to rename the repo, re-run checks, and then merge.

(We can always rename it back)

hadley (approved):

machow (approved):

PR #619, opened 2022-05-25
Refresh actions + tests

machow:
TODO: list out the env vars we need to run all tests

### DIFF
--- a/.github/workflows/rstudio-connect.yaml
+++ b/.github/workflows/rstudio-connect.yaml
@@ -62,24 +62,6 @@ jobs:
         with:
           upload-snapshots: true
 
-      # - name: Check
-      #   env:
-      #     _R_CHECK_CRAN_INCOMING_REMOTE_: false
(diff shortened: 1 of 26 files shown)


juliasilge:
For the MS365 boards, I think I figured out what to do for now. If you as a contributor want to test the MS365 boards (so you and me, @machow ):

- Go to https://onedrive.live.com and initiate a free account
- Add `PINS_MS365_USE_PERSONAL="true"` to the project-level `.Renviron`

Then I as an individual can test. These tests are SLOOOOOOOOW FYI. I don't think it's likely that we will want to set up these OneDrive tests to run on CI in the short term; I think that running them locally is good enough.

juliasilge:
After that last change ⤴️ locally, I am skipping the legacy tests (which is fine -- I don't think we should prioritize the v0 tests moving forward) and then _only_ the Azure tests, which @machow you are working on.

hadley (approved):

PR #632, opened 2022-08-10
Add more older R versions to R CMD check

juliasilge:
Closes #582, I hope 🤞


juliasilge:
Looks like the only thing that still failed on older versions of R is the [testing for the legacy `board_packages()`](https://github.com/rstudio/pins-r/runs/7775742546?check_suite_focus=true), because of creating a factor in R < 4.0. I added some coercion to character there just to get it to pass.

Issue #637, opened 2022-08-17
pin_versions_prune removes the newest, not the oldest versions


tomsing1:
I think the `pin_versions_prune` function is removing the youngest, not the oldest versions of a pin when the `n` argument is set:

```r
library(pins)

board <- board_temp(versioned = TRUE)

board %>% pin_write(data.frame(x = 1:5), name = "df")
Sys.sleep(1)
board %>% pin_write(data.frame(x = 2:6), name = "df")
Sys.sleep(1)
board %>% pin_write(data.frame(x = 3:7), name = "df")
Sys.sleep(1)

pin_versions_prune(board, "df", 1)
```

tomsing1:
Looking at how the versions to keep are identified in the https://github.com/rstudio/pins-r/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/pin_versions.R#L110 function, it seems that the versions are ordered by the `created` POSIX timestamp. 

At least on my system, that field doesn't have enough precision to distinguish between versions written within a one minute interval (e.g. all of the `created` fields in my example are the same `16:41:00` even though I ensured that they were written at least 1s apart).

But the `version` field has precision down to the second and ordering by this field would return the expected result. Because the `version` is also used to uniquely identify a file, perhaps the `versions_keep` function should receive the `version` as input, not `created`?

tomsing1:
Actually, is there a reason that the time stamp does not have second-level precision?
It seems the `parse_8601_compact` and `parse_8601 ` functions don't contain the `%S` part of the format string. Maybe that should be added to avoid losing that information? 

https://github.com/rstudio/pins-r/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/meta.R#L51

```r
> as.POSIXct(strptime(x, "%Y%m%dT%H%M", tz = "UTC"))
[1] "2022-08-18 00:05:00 UTC"
```
could be
```r
> as.POSIXct(strptime(x, "%Y%m%dT%H%M%S", tz = "UTC"))
[1] "2022-08-18 00:05:58 UTC"
```

That would make ordering versions by the `created` field work at least for versions created within > 1s of each other?

tomsing1:
Interesting - as I was writing more of my unit tests, I noticed that the `pin_write` function behaves differently when it is run within or outside unit tests (of the testthat variety). That's because the time stamp part of the `version`  is fixed in a unit test:

[...]

Issue #617, opened 2022-05-15
`created` in dataframe generated by `pin_versions()` doesn't update


juliasilge:
For `board_temp()`, the datetime `created` variable doesn't seem to track along with `version` and is always the same:

``` r
library(pins)
b <- board_temp(versioned = TRUE)
b %>% pin_write(head(mtcars), "mtcars")
#> Guessing `type = 'rds'`
#> Creating new version '20220515T194613Z-d58f7'
#> Writing to pin 'mtcars'
Sys.sleep(5)
b %>% pin_write(tail(mtcars), "mtcars")
#> Guessing `type = 'rds'`
#> Creating new version '20220515T194618Z-98d5a'
#> Writing to pin 'mtcars'
b %>% pin_versions("mtcars")
#> # A tibble: 2 × 3
#>   version                created             hash 
#>   <chr>                  <dttm>              <chr>
#> 1 20220515T194613Z-d58f7 2022-05-15 13:46:00 d58f7
#> 2 20220515T194618Z-98d5a 2022-05-15 13:46:00 98d5a
```

<sup>Created on 2022-05-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I am running into this for [vetiver docs](https://juliasilge.github.io/vetiver.dev/get-started/version.html#store-and-version-your-model), and you can also see it in the [pins docs](https://pins.rstudio.com/articles/pins.html#versioning).

juliasilge:
This is now fixed via #642 

``` r
library(pins)
b <- board_temp(versioned = TRUE)
b %>% pin_write(head(mtcars), "mtcars")
#> Guessing `type = 'rds'`
#> Creating new version '20220902T174120Z-209d4'
#> Writing to pin 'mtcars'
Sys.sleep(5)
b %>% pin_write(tail(mtcars), "mtcars")
#> Guessing `type = 'rds'`
#> Creating new version '20220902T174125Z-3fd04'
#> Writing to pin 'mtcars'
b %>% pin_versions("mtcars")
#> # A tibble: 2 × 3
#>   version                created             hash 
#>   <chr>                  <dttm>              <chr>
#> 1 20220902T174120Z-209d4 2022-09-02 11:41:20 209d4
#> 2 20220902T174125Z-3fd04 2022-09-02 11:41:25 3fd04
```

<sup>Created on 2022-09-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #642, opened 2022-08-22
Added %S to the datetime format string to avoid loss of precision

tomsing1:
- The `parse_8601_compact` function seems to be dropping the seconds from a timestamp. Adding the `%S` part of the format string preserves them. This might fix issue #637  - at least ensuring that files written within > 1s receive a unique version number.


juliasilge:
Closes #617

juliasilge (approved):
Thank you so much for this PR! Let's leave open #637 to create a better approach for `pin_versions()`.

PR #646, opened 2022-08-26
Move arrow to suggests.

### LINKED ISSUE #644: arrow to Suggests?
If you're reading and writing json or rds, `{arrow}` can be a relatively heavy, unused dependency. Would it be possible to move it to suggests? I'm not sure how tricky that might be (for example, to anticipate what types might need to be read), so I understand if this isn't feasible. I'm just seeing what I can cut out to make my containers a little smaller!

jonthegeek:
Closes #644.

Note: writing to arrow was previously returning the object (invisibly), rather than the fully qualified path (invisibly) like the help says. It now returns the path. As far as I can tell, no tests check this. I think json and csv likely return the object, but I haven't dug in to confirm.

### DIFF
--- a/tests/testthat/test-pin-read-write.R
+++ b/tests/testthat/test-pin-read-write.R
@@ -1,5 +1,6 @@
 test_that("can round trip all types", {
   skip_if_not_installed("qs")
+  skip_if_not_installed("arrow")
   board <- board_temp()
 
   # Data frames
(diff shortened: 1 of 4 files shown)


tests/testthat/test-pin-read-write.R
```diff
@@ -1,5 +1,6 @@
 test_that("can round trip all types", {
   skip_if_not_installed("qs")
+  skip_if_not_installed("arrow")
```
juliasilge, line 3: We didn't find this with `R-CMD-check-hard` because it's already skipping the test (qs not installed). Just a limitation of that current check. 🤷‍♀️

juliasilge (approved):
Thank you so much @jonthegeek! 🙌

Issue #625, opened 2022-07-07
board_s3() expose ServerSideEncryption and SSEKMSKeyId parameters


michkam89:
Hi, is it possible to use `board_s3()` with encrypted buckets? I couldn't find a way to pass `ServerSideEncryption` and `SSEKMSKeyId` parameters. 

Recently similar issue was for `targets` package [here](https://github.com/ropensci/targets/issues/855), it would be great to have similar solution here :)

juliasilge:
Related to #628 

If we pass the dots through to the paws `put_object` method, this would support `ServerSideEncryption` and `SSEKMSKeyId`. It would look like:

```r
board_sales <- board_s3("company-pins", prefix = "sales/")
board_sales %>% pin_write(mtcars, ServerSideEncryption = "AES256", SSEKMSKeyId = "your-key-here")
```

juliasilge:
Closed in #648 

Let us know if you have trouble with this moving forward!

fh-mthomson:
Wow, what great timing! I was just search for how to specify encryption via `board_s3()` and `pin_write()` in order to move to a `pins`-based workflow away from our current `aws.s3` functions. Thank you @juliasilge for this!

Unfortunately I'm hitting issues when trying to run the example above after installing the version on `main` at 7379792150e6b99cc3704f93e585c085aee67252:
``` r
board %>% pin_write(mtcars, ServerSideEncryption = "AES256")
board %>% pin_write(mtcars, Tagging = "key1=value1&key2=value2")
```

Both yield analogous error about unused dots:
```
Using `name = 'mtcars'`
Guessing `type = 'rds'`
Creating new version '20220909T233142Z-b8faa'
Error: AccessDenied (HTTP 403). Access Denied
Error in `pin_store()`:
! Arguments in `...` must be used.
✖ Problematic argument:
• ServerSideEncryption = "AES256"
Run `rlang::last_error()` to see where the error occurred.
Error in `pin_store()`:
! Arguments in `...` must be used.
✖ Problematic argument:
• ServerSideEncryption = "AES256"
Run `rlang::last_error()` to see where the error occurred.
Error in `pins::pin_write()`:
! Arguments in `...` must be used.
✖ Problematic argument:
• ServerSideEncryption = "AES256"
Run `rlang::last_error()` to see where the error occurred.
```

More detailed trace:
```
> rlang::last_error()
<error/rlib_error_dots_unused>
Error in `pins::pin_write()`:
[...]

PR #648, opened 2022-08-31
Pass the dots from `pin_write()` through to `s3_upload_file()`

### LINKED ISSUE #628: Usage of S3 tags with pins?
**Question**: Is there any way to upload objects to an S3 board with tags? If this is not possible, can it be added as a feature?

**Use Case**: My organization currently uses the {paws} package to upload shared datasets to an S3 bucket. This data is then used by various team members. However, it is also then read and ingested into other systems automatically based on what tags the object has. We would like to move to using the {pins} package for managing these shared datasets, but would need to be able to retain this ability to write tags to objects as they're being uploaded.

**Additional Notes**:
The `put_object` method from the {paws} package that {pins} relies on [allows tags to be passed along with uploaded objects](https://paws-r.github.io/docs/s3/put_object.html). I took a look at the `pins_write` function and, while it does take additional parameters, none of these arguments seem to be passed on to the `s3_upload_file` helper function:
https://github.com/rstudio/pins/blob/baaa304f876ec9cc98cc90d3333db157e34028f7/R/board_s3.R#L210-L220
https://github.com/rstudio/pins/blob/baaa304f876ec9cc98cc90d3333db157e34028f7/R/board_s3.R#L270-L277

juliasilge:
Closes #628 

``` r
library(pins)
b <- board_s3(
  "pins-test-hadley", 
  region = "us-east-2",
  access_key = Sys.getenv("PINS_AWS_ACCESS_KEY"),
  secret_access_key = Sys.getenv("PINS_AWS_SECRET_ACCESS_KEY")
)
b %>% pin_write(mtcars, Tagging = "key1=value1&key2=value2")
#> Using `name = 'mtcars'`
#> Guessing `type = 'rds'`
#> Creating new version '20220831T215459Z-66143'
#> Writing to pin 'mtcars'
```

<sup>Created on 2022-08-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/board_s3.R
+++ b/R/board_s3.R
@@ -275,6 +275,7 @@ s3_upload_yaml <- function(board, key, yaml) {
 }
 
 s3_upload_file <- function(board, key, path, ...) {
+  ellipsis::check_dots_used()
   body <- readBin(path, "raw", file.size(path))
   board$svc$put_object(
     Bucket = board$bucket,
--- a/R/board_s3.R
+++ b/R/board_s3.R
@@ -39,10 +39,12 @@
 #' an EC2 instance, using a role from another profile, and using multifactor
 #' authentication.
 #'
-#' # Caveats
+#' # Details
 #'
 #' * If you point at a bucket that's not created by pins, some functions
 #'   like `pins_list()` will work, but won't return useful output.
+#' * You can pass arguments for [paws.storage::s3_put_object] such as `Tagging`
+#'   and `ServerSideEncryption` through the dots of `pin_write()`.
 #'
 #' @inheritParams new_board
 #' @param bucket Bucket name.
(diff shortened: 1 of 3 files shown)


juliasilge:
Also closes #625

R/board_s3.R
```diff
@@ -209,14 +215,15 @@ pin_fetch.pins_board_s3 <- function(board, name, version = NULL, ...) {
 
 #' @export
 pin_store.pins_board_s3 <- function(board, name, paths, metadata,
-                                    versioned = NULL, ...) {
+                                    versioned = NULL, x = NULL, ...) {
```
juliasilge, line 218: The function signature here needed to be made the same as other `pin_store` methods.

machow (approved):
This behavior LGTM! I didn't try it locally (but am happy to if useful!)

Issue #627, opened 2022-07-12
board_url() can't download pins from boards created with board_folder()


innir:
Hi,

looking at the the [board_url()](https://pins.rstudio.com/reference/board_url.html) documentation

> A named character vector of URLs If the URL ends in a /, board_url will look for a data.txt that provides metadata. The easiest way to generate this file is to upload a pin directory created by [board_folder()](https://pins.rstudio.com/reference/board_folder.html).

it seems to be easy to create a board locally and write a pin to it with

```
> library(pins)
> board <- board_folder("test")
> board %>% pin_write(mtcars)
Using `name = 'mtcars'`
Guessing `type = 'rds'`
Creating new version '20220712T143218Z-66143'
Writing to pin 'mtcars'
> board$versioned
[1] FALSE
> fs::dir_tree("test")
test
└── mtcars
    └── 20220712T143218Z-66143
        ├── data.txt
        └── mtcars.rds
```

Now one should be able to just copy the folder `test` to some http server and get the pin `mtcars` back like

```
board <- board_url(c(mtcars = "https://example.com/test/mtcars/"))
board %>% pin_read("mtcars")
```

but the directory tree and the code at https://github.com/rstudio/pins/blob/main/R/board_url.R#L84-L90 makes it obvious that this will (and does) not work. The code expects `data.txt` at `https://example.com/test/mtcars/data.txt` while it is at `https://example.com/test/mtcars/20220712T143218Z-66143/data.txt`.

machow:
Hey--I wonder if this could be clarified in the documentation, but from what I can tell this phrase...

> upload a **pin directory** created by [board_folder()](https://pins.rstudio.com/reference/board_folder.html). [emph added]

Should not be referring to the board itself, or a pin directory, but a **pin version directory**. If you were to paste that version folder (`'20220712T143218Z-66143'`) somewhere, then you could use board_url to refer to it.

I'm not sure there's a simple way to support broader board functionality using `board_url`, without a way to list folders (which is not impossible with HTTP, but also dependent on HTTP servers doing things in a consistent way.)

Could you say more about the case you're trying to use it for?

innir:
True, one could use the pin version but that would draw it rather useless, as every update of the pin would force the reader of that pin to change the name :-/
For me the question it more, why does board_folder("test", versioned = **FALSE**) add a versioned directory?

[...]

PR #649, opened 2022-09-06
add custom type for `pin_write()`

### LINKED ISSUE #631: could `pin_write()` support `type = "arrow"` uncompressed?
This may also impact pins for Python, but I thought I would start here.

The motivation for this question is that [Arrow for JavaScript does not support compression](https://stackoverflow.com/a/66696736). I know there's a workaround using `pin_upload()` and `pin_download()`, but it seems to add friction for R/Python users.

To be clear, I'm not suggesting that any default behavior change; just to ask about the possibility for something like a `compression = FALSE` argument somewhere.

Here's a larger [writeup describing my motivation](https://observablehq.com/@ijlyttle/using-pins-with-javascript), with a first hack on how a `BoardUrl` could work using JavaScript.

Thanks!

ijlyttle:
fix #631

Sorry for delay, I finally got enough things pushed to the side of my "plate" so that I can devote some time to this - this draft is just to indicate I have not forgotten this or #627 .


ijlyttle:
I played around with a few ideas today, hoping each time to arrive at something simpler.

I still need to add tests, and more-complete documentation, but I wanted to run this proposed approach past you:

I am impressed with what you can do using `rlang::call_modify()`, but (at least to my eye) a "plain" function seems more familiar. You could call using:

- a bare-function: `.f = readr::write_csv` 
- a purrr-style anonymous function: `.f = ~arrow::write_feather(.x, .y, compression = "uncompressed")`

It has to be a function that takes the object first, then the path second.

The larger question seems to be the interaction between `type` and `.f`; I propose these possibilities for each:

`type`: 
  - `NULL`
  - one of the "standard" types (e.g. `"csv"`, `"json"`)
  - a string that begins with `"custom-"`.
  
`.f`: 
  - `NULL` 
  - a function

Here's what I propose for behavior:

| `type` | `.f` | behavior |
| ---  | --- | ---- |
| `NULL` | `NULL` | no change, use `guess_type()` |
| `NULL` | function | use `guess_type()`, inform that custom function is used, write using custom function |
| standard | `NULL` |  no change, write using `object_write()` |
| standard | function |  inform that custom function is used, write using custom function |
| custom | `NULL` |  abort in `object_write()` |
| custom | function | write using custom function (don't inform) |

Notes:

- This would let a user write `type = "arrow"` type uncompressed, or `type = "csv"` using readr or vroom. It becomes the users' responsibility ensure their custom funciton comports with the `type`. We would inform that a standard type is using a custom function.
- If you are specifying a custom type, it has to begin with `"custom-"`. This seems like a way to avoid collisions with future "standard" types.

---

I'll start working on tests, etc., but wanted to give you an early chance to react. Thanks!

juliasilge:
I think this is looking promising, but that table you made had me realize that we are probably bumping up against [argument dependencies/interdependence](https://design.tidyverse.org/args-independence.html). This has absolutely been a headache for me before 😩 so we definitely want to avoid this pattern if possible.

What do you think about adding support for passing a function to `type`, for folks to do something more custom? This is how [`unnest_tokens()`](https://github.com/juliasilge/tidytext/blob/main/R/unnest_tokens.R) works and it has been pretty decent for maintenance and flexibility. (Take a look at the `find_function()` function.)

ijlyttle:
I am with you. I now appreciate the danger of argument dependencies, thanks for helping me learn new things!

I will be happy to revise things to use the `type` argument. There are a couple things I would like to think about, in the fullness of time:

- If a function is passed to `type`, it seems the only possible value for `meta$type` is something like `"custom"`. This would seem to preclude being able to assert that something is a `"csv"`, but written using a different function. To be clear, I don't think this merits discarding the approach of overloading `type`, but it remains something I'd like to ponder the possibility to address.

- `pin_read()` could use an argument for a custom-reader function; presumably, this would be a separate issue.

ijlyttle:
Pondering this a bit more, there already exists the `"file"` type used with `pin_upload()` and `pin_download()`.

Overloading the `type` argument could provide an easier way to use `pin_upload()`, but instead of modifying `pin_read()`, one could:

```r
pin_download(my_board, "my-custom-pin") |>
  custom_reader_function()
```

ijlyttle:
I have run into another problem: how to specify the file extension. In `pin_write()`, this is handled by the `type`; if `type` is a function, we can't deduce the extension.

One possible solution might be to offer a helper function, `pin_make_writer()` (which could have a better name):

```r
#' @param f `function` that takes the object and a path, then writes the object to the path.
#' @param extension `character` extension for the file to be written.
#'
pin_make_writer <- function(f, extension) {

  function(x, name, envir = parent.frame()) {
    # determine path for file
    path <- fs::path_temp(fs::path_ext_set(fs::path_file(name), extension))

    # write object to file
    f(x, path)

    # delete file when environment that called writer is destroyed
    withr::defer(fs::file_delete(path), envir = envir)

    invisible(path)
  }
}
```

```r
arrow_pin_writer <- pin_make_writer(
  \(x, path) arrow::write_feather(x, path, compression = "uncompressed"), 
  extenstion = "arrow"
)

pin_write(my_board, mtcars, name = "mtcars", type = arrow_pin_writer)
```


I am sorry that I have turned this into a ride on the "API merry-go-round". While I enjoy an API discussion as much (perhaps more) than the next person, I know you all have other things to do.

I'll start coding in this direction unless or until I hear otherwise :)

ijlyttle:
I think I have something minimally working; here's what I get:

``` r
library("pins")

arrow_pin_writer <- pin_file_writer(
  ~arrow::write_feather(.x, .y, compression = "uncompressed"),
  extension = "arrow"
)

board <- board_temp()

pin_write(board, mtcars, type = arrow_pin_writer)
#> Using `name = 'mtcars'`
#> Creating new version '20220914T202009Z-8a125'
#> Writing to pin 'mtcars'

pin_upload(board, arrow_pin_writer(mtcars))
#> Guessing `name = 'mtcars.arrow'`
#> Creating new version '20220914T202009Z-8a125'

pin_download(board, "mtcars") |> arrow::read_feather() |> head()
#>    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> 1 21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> 2 21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> 3 22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> 4 21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> 5 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
#> 6 18.1   6  225 105 2.76 3.460 20.22  1  0    3    1

pin_download(board, "mtcars.arrow") |> arrow::read_feather() |> head()
#>    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> 1 21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> 2 21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> 3 22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> 4 21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> 5 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
#> 6 18.1   6  225 105 2.76 3.460 20.22  1  0    3    1
```

<sup>Created on 2022-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Assuming things are OK, some questions:

- do you wish to support using both `pin_write()` via `type` and `pin_upload()`?
- using `pin_upload()`, the filename is used as the default `name`, whereas `pin_write()` uses the filename sans extension. Is this difference intended, and if not, is it worth harmonizing (presumably, using a separate issue)?
- using `pin_upload()`, we don't have the message `Writing to pin '...'` - same question as above, is this worth harmonizing?

Thanks!

juliasilge:
So.... I am not feeling great about the `pin_file_writer()` approach; I think that is going to be confusing for many users. Maybe the nice interface of `pin_write()` isn't easily extensible to this kind of custom writing, because of needing to know how to write, plus what to write (i.e. the extension), etc.

**HOWEVER**, I am really excited about making this work better via `pin_upload()`! This may be the right place to provide more functionality/documentation for folks who want to pin using more custom options.

This fills the need, if I am understanding correctly?

``` r
library(pins)

b <- board_temp()
pin_name <- "my-mtcars"
path <- fs::path_temp(fs::path_ext_set(pin_name, "arrow"))
arrow::write_feather(mtcars, path, compression = "uncompressed")
b %>% pin_upload(path, pin_name, title = "My very excellent cars")
#> Creating new version '20220921T174004Z-8a125'
```

<sup>Created on 2022-09-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

What do you think we could do to make this a nicer user experience?

ijlyttle:
I think the `pin_upload()` approach is the simplest way to go. As you point out in the example, I think it becomes a matter of documentation, rather than adding a new function. That being the case, I'm thinking to close this PR, then propose another for documentation.

juliasilge:
That would be perfect @ijlyttle! Thank you for being willing to work through this in the way we did.

PR #652, opened 2022-09-10
add ... to s3_upload_yaml to fix issue noted in #625

fh-mthomson:
Fixes issue discussed in https://github.com/rstudio/pins-r/issues/625#issuecomment-1242558838

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,9 @@
 * Increased datetime precision to the second, for `pin_versions()` and related
   functions (#642, @tomsing1).
   
-* Pass the dots from `pin_write()` through to `s3_upload_file()` to support
-  S3 tagging, encryption options, etc for pins (#648).  
+* Pass the dots from `pin_write()` through to `s3_upload_file()` and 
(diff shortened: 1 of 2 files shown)


juliasilge:
Thank you so much for this @fh-mthomson! 🙌 

Can I make sure I am understanding this super clearly? This change will make it so both the _main_ file (such as `.rds` or `.csv`) and the `data.txt` are uploaded with the same arguments/options for [S3 PutObject](https://paws-r.github.io/docs/s3/put_object.html)? That seems like the right behavior to me, and it's hard to imagine a situation where that isn't what folks want.

fh-mthomson:
Exactly! Since the bucket to which I'm writing has SSE enabled, trying to write the `data.txt` file resulted in permissions errors when trying to call `put_object()` but without the `ServerSideEncryption` specification available. 

So, I think it's reasonable to expect that S3-level options should be passed through to all instances of `put_object()` since each call is inherently within the same bucket/prefix.

juliasilge (approved):
That makes sense! Thank you so much 🙏

PR #658, opened 2022-10-10
add documentation for custom formats

### LINKED ISSUE #631: could `pin_write()` support `type = "arrow"` uncompressed?

ijlyttle:
fix #631

Hi again,

Sorry I got pulled away on some other things...

I propose a short vignette, as it seemed a bit long for function documentation - happy to modify as you see fit!

Thanks!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pins (development version)
 
-* Added vignette to describe how to manage custom formats (#631, @ijlyttle).
+* Added vignette describing how to manage custom formats (#631, @ijlyttle).
 
 * Improved error message for `pin_versions()` (#657).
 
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
This is so great @ijlyttle! Thank you so much 🙌

ijlyttle:
My pleasure, thanks for catching the things I should have :)

PR #659, opened 2022-10-20
Document that pins doesn't create a new S3 bucket

### LINKED ISSUE #651: `board_s3()` will not create a bucket
We should address `board_s3()` not creating buckets, probably via documentation ("you can only write to an existing bucket") or possible by creating the bucket if it doesn't exist.

_Originally posted by @juliasilge in https://github.com/rstudio/pins-r/issues/608#issuecomment-1241360790_

juliasilge:
Closes #651

### DIFF
--- a/R/board_s3.R
+++ b/R/board_s3.R
@@ -41,13 +41,15 @@
 #'
 #' # Details
 #'
-#' * If you point at a bucket that's not created by pins, some functions
-#'   like `pins_list()` will work, but won't return useful output.
+#' * The functions in pins do not create a new bucket. You can create
+#'   a new bucket from R with [paws][paws.storage::s3_create_bucket].
+#' * Some functions like `pins_list()` will work for an S3 board, but don't
+#'   return useful output.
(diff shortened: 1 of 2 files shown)


R/board_s3.R
```diff
@@ -41,13 +41,15 @@
 #'
 #' # Details
 #'
-#' * If you point at a bucket that's not created by pins, some functions
-#'   like `pins_list()` will work, but won't return useful output.
+#' * The functions in pins do not create a new bucket. You can create
+#'   a new bucket from R with [paws][paws.storage::s3_create_bucket].
+#' * Some functions like `pins_list()` will work for an S3 board, but don't
```
machow, line 46: Should pins_list be pin_list? (super tiny nit!)

machow (approved):
LGTM!

R/board_s3.R
juliasilge: Thanks! And how long has that typo been there, I wonder??? 🙈

Issue #635, opened 2022-08-12
Investigate removing content-cache.yml, or refreshing it more frequently


machow:
For RSConnect boards, there is a `content-cache.yml` file that maps content name (e.g. `my-mtcars-pin`) to its underlying RSConnect identifier (a content guid). However, because changes to the mapping between content and guid can come from multiple places (e.g. other laptops, the web UI) this cache can easily go stale.

Opening this issue to track content-cache related problems, and to track research into what the effects of removing `content-cache.yml` would be.

Here are some related issues:

* https://github.com/rstudio/pins-r/issues/604
* https://github.com/rstudio/pins-r/issues/609

Here are issues related to the other cache file, `user-cache.yml`:

* https://github.com/rstudio/pins-r/issues/594

**Edit**: just a quick note on the two caches to clarify this issue.

* user-cache.yml: each entry maps a `<rstudio-guid>` -> `<username>`
* content-cache.yml: each entry maps `<username>/<content-name>` -> `rstudio-guid` and `url`

machow:
Potential paths that become slower when cache is removed:

* pin_meta (anything triggering rsc_content_find), where many users have content with the exact same name. (see [this API call](https://github.com/rstudio/pins-r/blob/main/R/board_rsconnect.R#L403))
*

machow:
Last thing--if we tackle this pins-python issue of being able to use guids when looking up underlying paths, we can likely remove the cache and work out the fast implementations for looking things up w/o the cache across both libraries.

* https://github.com/rstudio/pins-python/issues/79

machow:
From what I can tell, there's no way to completely fix the slowdowns we'd hit by removing the cache, using the current RSC API.

The issue is that we have two options for going from <user_name>/<content_name>  -> content :

* (1) API GET `v1/content?name=<content_name>`
  - issue: can return multiple entries (e.g. if multiple have content called "mtcars").
  - issue: the content payload does not contain the owner name (only their guid), so we have to look it up to filter multiple entries.
  - unusable solution: this route accepts a user `<guid>` parameter, but we only know `<user_name>`.
* (2) app endpoint GET `/applications?filter=content_type:pin&search=<content_name>`. 
  - benefit: this payload includes the names of content owners
  - issue: this returns any content that **contains** `<content_name>`, so there could be a lot more to filter.

[...]

PR #660, opened 2022-10-21
Use an R option or env variable to turn on/off Connect content cache

juliasilge:
Addresses #635


R/board_rsconnect.R
```diff
@@ -549,20 +550,33 @@ rsc_user_name <- function(board, guid) {
...
     list()
   }
 }
-update_cache <- function(path, key, value) {
-  cache <- read_cache(path)
-  cache[[key]] <- value
-  write_yaml(cache, path)
 
+update_cache <- function(path, key, value) {
+  if (use_cache()) {
```
juliasilge, line 561: I am checking the cache option in _both_ `read_cache()` and `update_cache()` even though the second calls the first so we avoid writing to the cache here in a couple of lines ⬇️

juliasilge:
This PR also does some code styling on `board_rsconnect.R`.

R/board_rsconnect.R
```diff
@@ -549,20 +550,33 @@ rsc_user_name <- function(board, guid) {
...
+    cache <- read_cache(path)
+    cache[[key]] <- value
+    write_yaml(cache, path)
+  }
```
juliasilge, line 565: I am checking the cache option in _both_ `read_cache()` and `update_cache()` even though the second calls the first so we avoid writing to the cache here

R/board_rsconnect.R
```diff
@@ -549,20 +550,40 @@ rsc_user_name <- function(board, guid) {
...
   value
 }
 
+clear_cache <- function(board,
+                        cache = c("content-cache.yml", "users-cache.yml")) {
+  path <- fs::path(board$cache, cache)
+  cli::cli_inform("Deleting {.file {path}}")
+  fs::file_delete(path)
+}
```
juliasilge, line 574: I added this new little function to delete the cache files. Right now, I have not exported or documented it, which makes it like a helper function for us to use in development and sort of secondarily, something for us to tell people to use when they get stuck. People would use it like:

```r
pins:::clear_cache(b, "content-cache.yml")
```
Thoughts? Should this be a regular exported function instead?

juliasilge: If we do really end removing these caches, then this function would need to be deprecated (if it were exported).

juliasilge: I decided to export and document it, and we can deal with that later.

R/board_rsconnect.R
```diff
@@ -549,20 +550,77 @@ rsc_user_name <- function(board, guid) {
...
+#' caches. You can manually delete a broken cache via `clear_connect_cache()`.
+#'
+#' @details The `use_connect_cache()` function uses the following method for
+#' detecting whether or not to use the content and user caches for Connect:
+#'
+#' 1. If the `pins_connect_cache` option is set to `FALSE`, `FALSE` is returned.
+#' 2. If the `PINS_CONNECT_CACHE` environment variable is set to `"false"`,
+#' `"FALSE"`, or `"False"`, `FALSE` is returned
+#' 3. In all other cases (including empty or `NULL` values for either of the
+#' above), `TRUE` is returned
```
juliasilge, line 592: The logic in the function is more complicated in order to give some helpful error messages for bad values. Although maybe I should reword this, in that case?

machow (approved):
LGTM!

juliasilge:
Closed by #667

Issue #640, opened 2022-08-20
Use of `is_testing()` is a problem for folks who program against pins functions


juliasilge:
Interesting - as I was writing more of my unit tests, I noticed that the `pin_write` function behaves differently when it is run within or outside unit tests (of the testthat variety). That's because the time stamp part of the `version`  is fixed in a unit test:

https://github.com/rstudio/pins-r/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/pin_versions.R#L146

I didn't expect that, and it took me a good while to figure out why my unit tests were failing. (I was writing the same file twice into a versioned board. My project includes tracking the metadata in a SQL database - and that complained about duplicated `version` keys.)

There is nothing intrinsically wrong with the current approach, but it reminds me a bit of the [Volkswagen scandal](https://en.wikipedia.org/wiki/Volkswagen_emissions_scandal) 😄 and the behavior definitely was surprising. (I solved it by temporarily unsetting the `TESTTHAT` environmental variable in my test.)

_Originally posted by @tomsing1 in https://github.com/rstudio/pins-r/issues/637#issuecomment-1220037880_

juliasilge:
The current use of `is_testing()`, for example, in `version_name()`:
https://github.com/rstudio/pins-r/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/pin_versions.R#L147-L148
makes it hard for folks to develop against pins functions and write their own tests.

juliasilge:
The new tests for `write_board_manifest()` (specifically the ones in `test-board_folder.R` that look at the structure of the manifest file) involve versions and could be improved in terms of how easy they are to grasp. Currently, we use `is_testing()` to handle versions in tests but we don't want to continue doing that.

One option for some cases might be to use snapshot testing with a function to redact versions.

juliasilge:
We could switch to use some mocking, like this:

``` r
library(pins)
b <- board_temp()

## pin_write() calls version_name() internally
b %>% pin_write(1:10, name = "nice-numbers", type = "json")
#> Creating new version '20230301T003852Z-c3943'
#> Writing to pin 'nice-numbers'
nice_meta <- b %>% pin_meta("nice-numbers")

library(mockery)

# call to version_name() is 2 layers down from pin_store()
stub(
  pin_store, 
  "version_name", 
  "20120304T050607Z-xxxxx",
  depth = 2
  )

## now with stub:
[...]

PR #661, opened 2022-10-26
Add new `write_board_manifest()` function

juliasilge:
Related to #650 

That PR is a bit sprawling so I think this will be better if we can break it up into some pieces. This PR implements _part_ of that change, adding the new function `board_manifest()` and the internal generic to call for various boards.

I have not implemented this for `board_rsconnect()`, because it's going to be pretty complicated. (Maybe look up all pins that the specific user can access?)

I changed the function name since this operates on a board and returns the board (it doesn't act on a pin).


juliasilge:
Currently one failure because we can't build the brand-new version of arrow on Windows.

@ijlyttle do you mind taking a look at this PR and seeing if you have feedback on the changes I made relative to your original work?

ijlyttle:
I think this looks great; the naming makes perfect sense!

Sorry I let this get away from me a bit. I agree that reducing the scope of each bit will make it much easier to manage.

Thanks!

machow (changes requested):
Producing a manifest seems super useful! I noticed we are writing a top-level file to a board, so mostly discuss that here, but wonder if it's worth separating out the creating of a manifest file, and this new behavior (a top-level file on a board) into separate PRs?

I'm a bit concerned that writing files (or even worse folders) that aren't pins to a board may produce surprising behaviors.

From a quick test, pins-python will see `pins.txt` as a pin (as will an R pins azure board?). This can be patched pretty quickly, but may take a bit of syncing across libraries / users sharing a board.

For the new behavior of adding a top-level file, can we test the following across all boards?:

* writing a `pins.txt` will not cause it to show up in pin_list
* writing a `pins.txt` will produce the correct result for pin_{exists,meta}("pins.txt")
* writing a `pins.txt` will not cause pins::cache_{info,prune} errors / surprising behavior
* explicate what will happen in pins legacy functions (e.g. could be explicitly unspecified)
* explicate / test what happens if a user attempts to create a pin named `"pins.txt"`
* explicate / test what happens if a user has created a pin named `"pins.txt"` and runs `board_manifest()`

It seems like these could go in `test_api_basic()` or a similarly named function.

Additional notes:

When I try it on s3, I get the warning below -- do you mind double checking?:

image

R/board_manifest.R
```diff
@@ -0,0 +1,80 @@
...
+    ~fs::path(.x, pin_versions(board, name = .x)$version) %>%
+      append_slash() %>%
+      as.list()
+  )
+  names(result) <- pin_names
+
+  result
+}
+
+append_slash <- function(x) {
```
ijlyttle, line 53: my bad - I think the warning may be because my helper function `append_slash()` is not vectorized.

Here's a better version:

```r
append_slash <- function(x) {
  has_slash <- grepl("/$", x)
  x[!has_slash] <- paste0(x[!has_slash], "/")
  
  x
}
```

R/board_manifest.R
```diff
@@ -0,0 +1,80 @@
...
+  if (!grepl("/$", x)) {
+    x <- paste0(x, "/")
+  }
+  x
+}
```
ijlyttle, line 58: my bad - I think the warning may be because my helper function `append_slash()` is not vectorized.

Here's a better version:

```r
append_slash <- function(x) {
  has_slash <- grepl("/$", x)
  x[!has_slash] <- paste0(x[!has_slash], "/")
  
  x
}
```

R/board_manifest.R
```diff
@@ -0,0 +1,80 @@
...
+  )
+  names(result) <- pin_names
+
+  result
+}
+
+append_slash <- function(x) {
+  if (!grepl("/$", x)) {
+    x <- paste0(x, "/")
+  }
```
ijlyttle, line 56: my bad - I think the warning may be because my helper function `append_slash()` is not vectorized.

Here's a better version:

```r
append_slash <- function(x) {
  has_slash <- grepl("/$", x)
  x[!has_slash] <- paste0(x[!has_slash], "/")
  
  x
}
```

ijlyttle: also, if it can be useful:

```r
expect_identical(append_slash("foo"), "foo/")
expect_identical(append_slash("foo/"), "foo/")
expect_identical(append_slash(c("foo/", "bar")), c("foo/", "bar/"))
expect_identical(append_slash(c("foo", "bar/")), c("foo/", "bar/"))
```

juliasilge: Thank you! 🙌

juliasilge:
I think I got all the tests and additional documentation added here. 

I don't think I'll open an issue about doing this for Connect until someone asks for it, as it is going to be a bear (just leave it non-implemented so default method gets called). As a reminder, this is the first ~half of what #650 aims for, and there will still be more to do for the overall workflow to be possible.

ijlyttle:
Hi @juliasilge - just a quick note, I have updated my workflows to use this branch, everything is working great with `_pins.yaml` and `write_board_manifest()`. 

This should not be in the least bit surprising, but figured you might like to know.

Thanks!

machow:
Sorry for the wait -- I'm going to double check quickly against pins python boards, but expect everything should work fine!

machow:
@juliasilge it appears adding a `pins.txt` file as a manifest breaks `pin_search` on python pins. The reason is that in the python version `pin_list()` returns the filename, and `pin_search` attempts to retrieve metadata for every pin as part of search.

Here's the relevant R pins [pins_search code fetching metadata](https://github.com/rstudio/pins-r/blob/b0b6a47f654a1b55159991aeb167dbfcbe6ea8a5/R/pin_search.R#L32-L34) (which works fine with the manifest file).

wdyt is a good move here? I can modify pins-python to ignore `pins.txt` files, but I'm wondering what the right way sequence is for releasing..

machow:
> This is helpful in general in case people are putting other files in, say, an S3 bucket that holds pins.

I wonder if we should discourage people from putting other files/folders into a pin board? Ignoring files will work okay, but if people put folders in they'll get listed as pins (afaict the exact behavior varies from board to board).

For example, if a folder with a file in it (`abc/xyz.txt`) is inside of an S3 board, then pin_search will raise an error (because it will try to pull the metadata for a pin named `abc`):

image

machow (approved):

R/board_azure.R
```diff
@@ -226,13 +245,11 @@ azure_ls <- function(board, dir = "") {
   paths <- AzureStor::list_storage_files(
     board$container,
     dir = dir,
-    recursive = FALSE,
-    info = "name"
+    recursive = FALSE
```
hadley, line 248: Why did this change?

R/board_ms365.R
```diff
@@ -95,11 +95,11 @@ board_ms365 <- function(drive, path, versioned = TRUE, cache = NULL, delete_by_i
 
   cache <- cache %||% board_cache_path(paste0("ms365-", hash(folder$properties$id)))
   new_board_v1("pins_board_ms365",
-    folder = folder,
-    path = path,
-    cache = cache,
-    versioned = versioned,
-    delete_by_item = delete_by_item
+               folder = folder,
```
hadley, line 98: Unnecessary change?

R/board.R
```diff
@@ -160,6 +160,95 @@ board_deparse.pins_board <- function(board, ...) {
...
+  # given board, return named list:
+  #   - names are pin names
+  #   - values are relative paths to version directories
+
+  pin_names <- pin_list(board)
+
+  result <- map(
+    pin_names,
+    ~fs::path(.x, pin_versions(board, name = .x)$version) %>%
+      append_slash() %>%
```
hadley, line 221: I think it's worth a brief comment explaining why this is needed

tests/testthat/test-board_folder.R
```diff
@@ -43,6 +44,27 @@ test_that("can deparse", {
...
+  write_board_manifest(b)
+
+  # names are correct
+  manifest <- yaml::read_yaml(fs::path(b$path, manifest_pin_yaml_filename))
+  expect_identical(names(manifest), pin_list(b))
+
+  # values (relative paths to versions) are correct
+  imap(
+    manifest,
+    ~ expect_identical(
```
hadley, line 60: Is there some way you could construct this expected values so the structure is obvious at a glance? And maybe worthwhile to create a pin with more than one version?

tests/testthat/test-utils.R
```diff
@@ -11,3 +11,10 @@ test_that("write_yaml can write non-UTF8 data", {
   y <- yaml::read_yaml(path)
   expect_equal(y, list("fa\u00e7ile" = "fa\u00e7ile"))
 })
+
+test_that("can append slash", {
```
hadley, line 15:
```suggestion
test_that("can append slash if needed", {
```

hadley (approved):

R/board_azure.R
juliasilge: Because we need to know `isdir`, not only the name; see line 250 ⤵️

tests/testthat/test-board_folder.R
juliasilge: Added another version to one of the test pins 👍 

I thought a while about a way to construct the expected values, but it is not obvious to me with the versions. We currently handle that with `is_testing()` in `version_name()` but that has caused problems as outlined in #640. I wonder if maybe the way to go is snapshot tests with a function to redact versions. I'll add that in the existing issue.

PR #662, opened 2022-10-26
Add test for writing with `board_folder()` ➡️ reading with `board_url()`

juliasilge:
Related to #650 

This particular test only checks putting the whole path including the version into the URL for `board_url()`, but let's get this piece reviewed and merged so we have a good strategy for testing this kind of process.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -32,39 +32,22 @@ jobs:
       R_KEEP_PKG_SOURCE: yes
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


juliasilge:
@ijlyttle Do you have any feedback for this testing strategy? We'll add more tests like this once we get the manifests added, and `board_url()` set up to use them.

ijlyttle:
If I understand right, the idea is to use webfakes as a substitute for rawgithub. 

This way, the tests can be included in a single pass, rather than pre-populating with a versioned pin for testing `board_url()`?

Works for me (and I need to read more about webfakes)!

machow (approved):
Ah, thanks -- webfakes looks super handy!

Issue #596, opened 2022-02-06
pin_write() automatically converts characters to factors 


erwinlares:
Hello fellow homo sapiens, 

pin_write() seems to be converting character variables to factors despite what I read here in issue #175 

I read in the Palmer Penguins data set from a csv file. Here's what the tibble looks like.
```
> penguins |> str()
spec_tbl_df [344 × 8] (S3: spec_tbl_df/tbl_df/tbl/data.frame)
 $ species          : chr [1:344] "Adelie" "Adelie" "Adelie" "Adelie" ...
 $ island           : chr [1:344] "Torgersen" "Torgersen" "Torgersen" "Torgersen" ...
 $ bill_length_mm   : num [1:344] 39.1 39.5 40.3 NA 36.7 39.3 38.9 39.2 34.1 42 ...
 $ bill_depth_mm    : num [1:344] 18.7 17.4 18 NA 19.3 20.6 17.8 19.6 18.1 20.2 ...
 $ flipper_length_mm: num [1:344] 181 186 195 NA 193 190 181 195 193 190 ...
 $ body_mass_g      : num [1:344] 3750 3800 3250 NA 3450 ...
 $ sex              : chr [1:344] "male" "female" "female" NA ...
 $ year             : num [1:344] 2007 2007 2007 2007 2007 ...
```

I pin it to my board 
```
board <- board_rsconnect()
board |> pin_write(penguins, "palmerpenguins",
                    type = "csv")
```

When I read it back, I get this with all character variables now as factors.

```
> pin_read(board, "lares/palmerpenguins") |> str()
'data.frame':	344 obs. of  8 variables:
 $ species          : Factor w/ 3 levels "Adelie","Chinstrap",..: 1 1 1 1 1 1 1 1 1 1 ...
 $ island           : Factor w/ 3 levels "Biscoe","Dream",..: 3 3 3 3 3 3 3 3 3 3 ...
 $ bill_length_mm   : num  39.1 39.5 40.3 NA 36.7 39.3 38.9 39.2 34.1 42 ...
 $ bill_depth_mm    : num  18.7 17.4 18 NA 19.3 20.6 17.8 19.6 18.1 20.2 ...
 $ flipper_length_mm: int  181 186 195 NA 193 190 181 195 193 190 ...
 $ body_mass_g      : int  3750 3800 3250 NA 3450 3650 3625 4675 3475 4250 ...
[...]

PR #664, opened 2022-10-27
No more `stringsAsFactors`

### LINKED ISSUE #626: Is `stringsAsFactors = TRUE` the best option when `type = "csv"`?
This has bitten me already multiple times:

https://github.com/rstudio/pins/blob/debc94529b8b73cda41ab50419d6276cd6c56d24/R/pin-read-write.R#L180

juliasilge:
Closes #626 

Setting `stringsAsFactors = TRUE` seems quite unexpected now, especially 2+ years after this default changed in base R. It results in behavior that surprises many (I would guess most) users, for example see #596.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 
 * Improved error message for `pin_versions()` (#657).
 
+* Changed `type = "csv"` to use R's default value for `stringsAsFactors` i.e. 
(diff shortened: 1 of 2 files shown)


juliasilge:
This is a breaking change, so let's bump to pins 1.1.0 at the next release.

PR #667, opened 2022-10-31
Use environments for Connect content + user caches

juliasilge:
Addresses #635 

Instead of removing these troublesome caches like #660 was headed for, this PR instead changes these caches over to use environments (instead of files on disk). This means that:

- caches will not persist between sessions
- caches will only get into a broken state if a big change is made with the server while a user is working with pins from R
- caches can be fixed by restarting R
- the first time functions are called they will be slow, but subsequent calls will have the same benefits of caching as we have now

### DIFF
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -545,21 +545,6 @@ rsc_user_name <- function(board, guid) {
   )
 }
 
-read_cache <- function(path) {
-  if (file.exists(path)) {
-    yaml::read_yaml(path, eval.expr = FALSE)
-  } else {
-    list()
-  }
-}
-update_cache <- function(path, key, value) {
-  cache <- read_cache(path)
-  cache[[key]] <- value
-  write_yaml(cache, path)
-
-  value
-}
-
 # helpers -----------------------------------------------------------------
 
 rsc_path <- function(board, path) {
--- a/R/board_rsconnect.R
+++ b/R/board_rsconnect.R
@@ -388,13 +388,14 @@ board_pin_find.pins_board_rsconnect <- function(board,
 
 # Content -----------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


R/board_rsconnect.R
```diff
@@ -388,17 +388,20 @@ board_pin_find.pins_board_rsconnect <- function(board,
 
 # Content -----------------------------------------------------------------
 
+rsconnect_content_cache_env <- rlang::new_environment()
+rsconnect_user_cache_env <- rlang::new_environment()
```
juliasilge, line 392: I put these here? And also I think it's simplest to still have two of them, one for the pieces of content and one for the users.

juliasilge:
This idea worked out great IMO @hadley!

R/board_rsconnect.R
hadley: I think this is a totally reasonable approach but we're starting to build a bit of a consensus around using [`the`](https://github.com/tidyverse/design/issues/126) as a sentinel for a mutate package environment. If you wanted to use that, I think you'd do:

```R
the <- rlang::new_environment()
the$content_cache <- rlang::new_environment()
the$user_cache <-  rlang::new_environment()
```

hadley (approved):

machow:
Just a heads up -- AFAICT one side-effect of this change is that if users restart their R sessions, they'll be unable to access cached RSC pin content without an internet connection. The issue is that already accessed content will be in their on-disk cache, but there will be no mapping back to it :/

(it might be worth letting people access pins via guid in the future to work around? see https://github.com/rstudio/pins-python/issues/79)

juliasilge:
@machow That is not different than if we turned the cache off via an environment variable or eventually removed the cache altogether, correct? Those were the other options we have been talking about.

I like the idea of using the GUID as an internal bit of metadata.

PR #668, opened 2022-11-01
Add docs for MinIO storage

### LINKED ISSUE #663: MinIO S3 storage
Would it be possible to implement connection to MinIO S3? The same way as the `board_s3` function works, but also for MinIO SR instead for AWS only.
I am using Dell ECS S3 storage (AWS competitor) and I can connect via MinIO. It would be awesome if I could use a board there.

juliasilge:
Closes #663 

@tomasjanikds do you mind checking this out and seeing if it would have helped you when you were trying to get this to work?

### DIFF
--- a/R/board_s3.R
+++ b/R/board_s3.R
@@ -1,7 +1,7 @@
 #' Use an S3 bucket as a board
 #'
-#' Pin data to a bucket on Amazon's S3 service, using the paws.storage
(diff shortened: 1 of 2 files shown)


tomasjanikds:
@juliasilge apologies for the delay. Yes the updated docs would help.

Issue #673, opened 2022-11-05
board_s3, Shiny deployment, and paws dependency


vgXhc:
I just walked through this issue with shinyapps.io support, and I think this qualifies as a {{pins}} issue:

I have a Shiny app that uses `board_s3` to read in data. Reprex (if you provide your own S3 credentials)

```
library(pins)
board <- board_s3("vzpins",
region = "us-east-1",
access_key = Sys.getenv("S3_ACCESS_KEY"),
secret_access_key = Sys.getenv("S3_SECRET_ACCESS_KEY"))

# Define UI for application that draws a histogram
ui <- fluidPage(

# Application title
titlePanel("Old Faithful Geyser Data"),

# Sidebar with a slider input for number of bins
sidebarLayout(
sidebarPanel(
sliderInput("bins",
"Number of bins:",
min = 1,
max = 50,
value = 30)
),

# Show a plot of the generated distribution
mainPanel(
plotOutput("distPlot")
)
)
)

# Define server logic required to draw a histogram
server <- function(input, output) {

output$distPlot <- renderPlot({
# generate bins based on input$bins from ui.R
x <- faithful[, 2]
bins <- seq(min(x), max(x), length.out = input$bins + 1)

# draw the histogram with the specified number of bins
hist(x, breaks = bins, col = 'darkgray', border = 'white')
})
}

# Run the application
shinyApp(ui = ui, server = server)
```

The app worked fine locally, but after deploying it on shinyapps.io I got a (not very helpful) error:

```
022-10-19T15:19:22.316043+00:00 shinyapps[6862336]: Error in value[[3L]](cond) :
2022-10-19T15:19:22.316075+00:00 shinyapps[6862336]: Calls: local ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
2022-10-19T15:19:22.316080+00:00 shinyapps[6862336]: Execution halted
2022-10-19T15:19:22.316096+00:00 shinyapps[6862336]: Shiny application exiting ...
```

Long story short, we ultimately identified the issue: When deploying the app with `rsconnect()`, only imported package dependencies are  included in the bundle; suggested packages are not. `board_s3` relies on `paws.storage`, which is apparently not a required package. Including `library(paws.storage)` in the code of the app made it work:

`````
library(pins)
library(paws.storage)
board <- board_s3("vzpins",
region = "us-east-1",
access_key = Sys.getenv("S3_ACCESS_KEY"),
[...]

PR #674, opened 2022-11-09
Add `required_pkgs()` methods for boards

juliasilge:
This is related to #673 as well as rstudio/vetiver-r#149.

We can use this functionality for creating an `renv.lock` or Dockerfile, and also to generate some more clear documentation.


R/testthat.R
```diff
@@ -111,6 +111,10 @@ test_api_basic <- function(board) {
     )
   })
 
+  testthat::test_that("can find board required pkgs", {
+    testthat::expect_snapshot(required_pkgs(board))
```
juliasilge, line 115: I think a snapshot test is probably the best way to check these new methods across all the boards?

R/board_azure.R
```diff
@@ -194,6 +194,12 @@ board_deparse.pins_board_azure <- function(board, ...) {
   expr(board_azure(!!container, path = !!board$path))
 }
 
+#' @rdname required_pkgs.pins_board
+#' @export
+required_pkgs.pins_board_azure <- function(x, ...) {
+  ellipsis::check_dots_empty()
+  "AzureStor"
```
hadley, line 201: It slightly concerns me that we're now recording this information in two places so it might get out of sync. But I don't see any way to avoid that.

hadley (approved):

PR #677, opened 2022-11-16
Add `tags` to metadata

### LINKED ISSUE #570: Add tags field to metadata

juliasilge:
Closes #570 

This PR adds a `tags` field to pin metadata. It is used like this:

``` r
library(pins)
b <- board_rsconnect()
#> Connecting to RSC 2022.09.0 at <https://colorado.rstudio.com/rsc>
b %>% pin_write(
  1:10, 
  "julia.silge/great-numbers", 
  tags = c("i-love-integers", "numbers-are-fun")
)
#> Guessing `type = 'rds'`
#> Writing to pin 'julia.silge/great-numbers'
```

<sup>Created on 2022-11-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

You can see how this is [stored on Connect here](https://colorado.rstudio.com/rsc/content/cc8ef31c-d6b3-4e0a-be8e-7992efffaee0/) (click "Raw metadata" to see the tags stored in YAML).

This adds a new argument in `pin_write()` between `description` and `metadata`, which could cause problems for any folks who have used `metadata` by position only. It does not fail silently, though:

``` r
library(pins)
b <- board_temp()
b %>% pin_write(1:10, "great-numbers", "json", "My beautiful numbers", "Some numbers for you to enjoy", list(a = "a"))
#> Error in `check_tags()` at pins-r/R/pin-read-write.R:88:2:
#> ! `tags` must be a character vector

#> Backtrace:
#>     ▆
#>  1. ├─b %>% ...
#>  2. └─pins::pin_write(...)
#>  3.   └─pins:::check_tags(tags) at pins-r/R/pin-read-write.R:88:2
#>  4.     └─rlang::abort("`tags` must be a character vector") at pins-r/R/pin-read-write.R:256:4
```

<sup>Created on 2022-11-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think the number of folks who have used `metadata` like this (by position) must be very small, though, because there are so many arguments before it. Is this a breaking change? Probably?

For future extensions, on Connect we could [try to store these `tags` as Connect tags](https://docs.posit.co/connect/cookbook/organizing/#tags). Publishers cannot create tags, so the plan would be to check if the tag exists and then add it to the pin.

### DIFF
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -52,6 +52,8 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   others can understand what the pin contains. If omitted, a brief
 #'   description of the contents will be automatically generated.
 #' @param description A detailed description of the pin contents.
+#' @param tags A character vector of tags for the pin; most important for
+#'   discoverability on shared boards.
 #' @param metadata A list containing additional metadata to store with the pin.
 #'   When retrieving the pin, this will be stored in the `user` key, to
 #'   avoid potential clashes with the metadata that pins itself uses.
@@ -67,6 +69,7 @@ pin_write <- function(board, x,
                       type = NULL,
                       title = NULL,
                       description = NULL,
+                      tags = NULL,
                       metadata = NULL,
                       versioned = NULL,
                       ...) {
@@ -82,6 +85,7 @@ pin_write <- function(board, x,
       abort("Must supply `name` when `x` is an expression")
     }
   }
+  check_tags(tags)
   check_metadata(metadata)
   if (!is_string(name)) {
     abort("`name` must be a string")
@@ -100,6 +104,7 @@ pin_write <- function(board, x,
     paths = path,
     type = type,
     title = title %||% default_title(name, data = x),
+    tags = tags,
     description = description
   )
   meta$user <- metadata
@@ -244,6 +249,11 @@ check_board <- function(x, v1, v0) {
 check_name <- function(x) {
   if (grepl("\\\\|/", x, perl = TRUE)) {
     abort("`name` must not contain slashes", class = "pins_check_name")
+  }
+}
+check_tags <- function(x) {
+  if (!is.null(x) && !is_character(x)) {
+    abort("`tags` must be a character vector")
   }
 }
 check_metadata <- function(x) {
--- a/R/meta.R
+++ b/R/meta.R
@@ -32,7 +32,11 @@ write_meta <- function(x, path) {
 
 # pin metadata ------------------------------------------------------------
 
-standard_meta <- function(paths, type, title = NULL, description = NULL) {
+standard_meta <- function(paths,
+                          type,
+                          title = NULL,
+                          description = NULL,
+                          tags = NULL) {
   list(
     file = fs::path_file(paths),
     file_size = as.integer(fs::file_size(paths)),
@@ -40,6 +44,7 @@ standard_meta <- function(paths, type, title = NULL, description = NULL) {
     type = type,
     title = title,
     description = description,
+    tags = tags,
     created = format(Sys.time(), "%Y%m%dT%H%M%SZ", tz = "UTC"),
     api_version = 1
   )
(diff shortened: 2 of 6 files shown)


juliasilge:
This change is backwards compatible for reading pins that were stored without `tags`:

``` r
library(pins)
b <- board_rsconnect()
#> Connecting to RSC 2022.09.0 at <https://colorado.rstudio.com/rsc>
b %>% pin_meta("julia.silge/wow-another-mtcars-model")
#> List of 11
#>  $ file       : chr "wow-another-mtcars-model.rds"
#>  $ file_size  : 'fs_bytes' int 5.07K
#>  $ pin_hash   : chr "ebc45bb5da89fbcc"
#>  $ type       : chr "rds"
#>  $ title      : chr "wow-another-mtcars-model: a pinned list"
#>  $ description: chr "An OLS linear regression model"
#>  $ created    : POSIXct[1:1], format: "2022-10-07 14:45:38"
#>  $ api_version: num 1
#>  $ user       : list()
#>  $ name       : chr "julia.silge/wow-another-mtcars-model"
#>  $ local      :List of 4
#>   ..$ dir       : 'fs_path' chr "~/Library/Caches/pins/rsc-e62371cfd77db754024f9c5ed3556a73/3ff71b31-f257-480a-a87b-fc6dc9a7e4dc/63410"
#>   ..$ url       : chr "https://colorado.rstudio.com/rsc/content/3ff71b31-f257-480a-a87b-fc6dc9a7e4dc/_rev63410/"
#>   ..$ version   : chr "63410"
#>   ..$ content_id: chr "3ff71b31-f257-480a-a87b-fc6dc9a7e4dc"
```

<sup>Created on 2022-11-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

juliasilge:
@machow do you mind taking a look, with an eye to a) whether this change would be a problem for pins for Python as it exists today and b) whether you think this approach would work for Python (later, when adding it may be appropriate)?

R/pin-read-write.R
```diff
@@ -67,6 +69,7 @@ pin_write <- function(board, x,
                       type = NULL,
                       title = NULL,
                       description = NULL,
+                      tags = NULL,
```
machow, line 72: I wonder if it might be helpful to move the tags argument to the back, to keep backwards compatibility? I'm not sure how often metadata and versioned get used, but the word "tags" is short enough that it should be quick to type!

machow (changes requested):
This LGTM, and should be pretty quick to add to python pins!

I left a quick question on whether tags should be added as the last argument to `pin_write`. It seems like "tags" would be pretty quick to type for kw args, and will let this be backwards compatible (but I get wanting it to come before rarely used arguments 😭)

R/pin-read-write.R
juliasilge: Yeah, after some time, I think this is the right call. 👍

juliasilge:
I think you are right @machow that we should just put this at the end. I still kept it before the dots `...` which means it could possibly still be a breaking change for folks who have used existing args by position only (not by name) and then passed something through the dots; I think this must be very rare since there are so many arguments ahead of the dots.

R/pin-read-write.R
```diff
@@ -60,6 +60,8 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   lists and rds for everything else.
 #' @param versioned Should the pin be versioned? The default, `NULL`, will
 #'   use the default for `board`
+#' @param tags A character vector of tags for the pin; most important for
+#'   discoverability on shared boards.
 #' @rdname pin_read
 #' @export
 pin_write <- function(board, x,
```
hadley, line 67: If I was to design this function today, I would put `...` after `x`.

R/pin-read-write.R
```diff
@@ -251,6 +256,11 @@ check_metadata <- function(x) {
     abort("`metadata` must be a list")
   }
 }
+check_tags <- function(x) {
+  if (!is.null(x) && !is_character(x)) {
+    abort("`tags` must be a character vector")
```
hadley, line 261:
```suggestion
    abort("`tags` must be a character vector.")
```

Ideally you'd say what it actually is too (using `friendly_type`), but if you start going to far down that route you might as well bundle it into a thorough review of argument checking and then use the latest helpers from rlang.

hadley (approved):

PR #692, opened 2022-12-16
vignette for `board_url()`

### LINKED ISSUE #685: Add new vignette about using manifest + `board_url()`
As discussed in #681, let's add a vignette on how to use a manifest with `board_url()`. Thank you so much for being willing to work on it @ijlyttle!

It is not a good fit for an example, because you need something like a server or webfakes to serve the directory with a manifest. Since there will be no example, let's be sure to _link_ to the vignette in the docs for both the manifest and `board_url()`.

ijlyttle:
fix #685

@juliasilge - just to let you know I'm working on this, and to give you an early opportunity to point me back to the right path, if need be.


juliasilge:
This looks good so far! I would probably do "Publishing" before "Consuming".

ijlyttle:
My first thought there would be more consumers than publishers, but I agree with you; it would make the programming a whole lot more convoluted.

ijlyttle:
@juliasilge - non-urgent question for you:

I'm failing the R-CMD-check-hard workflow because webfakes is not a hard dependency. I don't know if the right move is to make it a hard dependency, or instead to add it to the workflow's `extra-packages:`.

Thanks!

ijlyttle:
The answer was in the other vignette 😅

ijlyttle:
I think I've got this to a "discussable" point; the sections on publishing and consuming are ready for review.

There's an additional section on platforms. I wanted to create examples for Quarto and pkgdown, if for no other reason than to demonstrate clearly to myself that it works. A couple points here:

- I don't know if you will want to include the links in the vignette, as it adds a bit of risk: I could absent-mindedly delete the repos, or repurpose them to advertise sports-betting :). That being the case, I'd be happy to turn the repos over to you all, or find another way to accomplish the goal.
- I found something in Quarto that I'd like to clear up before merging this vignette: I don't quite understand how it handles external resources. I'm sure this won't take too long.

I also created a section for S3, but I don't have an S3 account. From what I read, it should just work™️ to publish your bucket as a website.

Thanks!

vignettes/using-board-url.Rmd
```diff
@@ -0,0 +1,208 @@
...
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  eval = rlang::is_installed("webfakes")
+)
+```
+
+The pins package supports back-and-forth collaboration using, for example, `board_s3()`.
+The goal of this vignette is to show how to publish a pins-board to a website, bringing your pins to a wider audience.
```
juliasilge, line 19:
```suggestion
The goal of this vignette is to show how to publish a board of pins to a website, bringing your pins to a wider audience.
```

vignettes/using-board-url.Rmd
```diff
@@ -0,0 +1,208 @@
...
+We make a quick check of our board to ensure we have one pin, named `"mtcars"`, with two versions:
+
+```{r board-pin-list}
+board %>% pin_list()
+
+board %>% pin_versions("mtcars")
+```
+
+Because a `board_url()` is consumed over the web, it doesn't have access to a file system the way, for example, a `board_folder()` has; our workaround is to create a manifest file.
+When a `board_url()` is created, it uses this file to discover the pins and their versions.
```
juliasilge, line 85:
```suggestion
When a `board_url()` is set up by a consumer for reading, the function uses this file to discover the pins and their versions.
```

vignettes/using-board-url.Rmd
```diff
@@ -0,0 +1,208 @@
...
+
+## Publishing platforms
+
+The goal of this section is to illustrate, for a few platforms, the particular way to publish a board as a part of a website.
+
+### Quarto
+
+Using Quarto, you can [publish](https://quarto.org/docs/publishing/) to GitHub Pages, Quarto Pub, and Netlify, among others.
+
+Let's say you set up a board using a directory called `"pins-board"`; of course, you may wish to call use a different name.
```
juliasilge, line 161:
```suggestion
Let's say you set up a board using a directory called `"pins-board"`; of course, you may wish to use a different name.
```

juliasilge (commented):
This is looking great! 🙌 I made a few minor suggestions in the narrative.

I see that we're waiting on something from the Quarto folks to get that squared away. 👍 

I see what you mean about those nice example repos you created. Hmmm... Let's keep them there for now, unless we decide there is something better suited for them.

ijlyttle:
Hi @juliasilge - thanks for your feedback and patience, we're getting there!

I agree that we should let the Quarto folks sort things out, and leave the demo repos where they are until a better solution presents itself.

Hopefully be back soon!

juliasilge:
@ijlyttle I think we want to move forward with a pins release fairly expediently, so my plan here is to remove the Quarto section of this vignette and open an issue to add that back in later, when the change is made in Quarto. I'll take a stab at finishing up this vignette!

ijlyttle:
Sounds good - hopefully everything gets worked out soon!

vignettes/using-board-url.Rmd
```diff
@@ -0,0 +1,208 @@
...
+  - Write your datasets to the board using `pin_write()`.
+  - At the end, call `write_board_manifest()`.
+
+The R Packages book offers this observation on CRAN and package data:
+
+> Generally, package data should be smaller than a megabyte - if it’s larger you’ll need to argue for an exemption.
+
+Publishing a board on your pkgdown site provides a way to offer extended versions of your data.
+
+Here's an example of a [toy package](https://github.com/ijlyttle/pkgpins) and its [pkgdown site](https://ijlyttle.github.io/pkgpins/).
```
juliasilge, line 203: I think I'm going to take this out of the vignette since we're not going to share a similar example for S3. However @ijlyttle would you be OK with keeping that repo + site up at least until we publish the next pins release blog post (plus a while after that), so I can link to them from there?

vignettes/using-board-url.Rmd
```diff
@@ -0,0 +1,208 @@
...
+    ```
+
+  - You may wish to add `/pins-board/` to your project's `.gitignore` file.
+
+Here's an example of a [Quarto repository](https://github.com/ijlyttle/quarto-pins) and its [published website](https://ijlyttle.github.io/quarto-pins/).
+
+### pkgdown
+
+Pins offers another way for R-package developers to share data associated with a package. 
+Publishing a package dataset as a pin can let you extend your data's "audience" to Python users.
```
juliasilge, line 184: I'm going to take out "Python users" here because Python's `board_url()` doesn't support the manifest yet.

vignettes/using-board-url.Rmd
ijlyttle: No worries - I'll keep the repo & site up

vignettes/using-board-url.Rmd
ijlyttle: In which case, I'll look forward to its return :)

juliasilge:
@ijlyttle Would you mind looking over my edits to see if any changes I made don't make sense? Give it one more look before I merge?

ijlyttle:
Hi @juliasilge, it looks great - much improved, in fact!

Thanks!

PR #681, opened 2022-12-03
Revamp `board_url()` to handle manifest files

juliasilge:
Related to #661 

Closes #650

PR #693, opened 2022-12-17
fix small typo

ijlyttle:
from #681


ijlyttle:
this will teach me not to start a PR using github.com, then continue it on my local machine :/

PR #695, opened 2022-12-20
Add new `board_gcs()` for Google Cloud Storage

### LINKED ISSUE #572: Implement modern google cloud storage board
Please 👍 this issue if you'd like this implemented.

juliasilge:
Closes #572


MarkEdmondson1234:
suppressMessages() should do it for the upload feedback. If it needs more fine grain control it could be conditional on options(googleAuthR.verbose = 4) - the default is 3, 4 will suppress all messages. 

The objects aren't really in folders more just names with / in them, so listing folders isn't directly supported. If folder structure is important perhaps make new buckets?

Prefix is only to filter results coming back that begin with x, so can do pseudo "list a folder" with "everything that starts with "folder/"

juliasilge:
@MarkEdmondson1234 it looks like `googleCloudStorageR::gcs_get_object()` does not respect `googleAuthR.verbose`:

``` r
withr::with_options(
  list(googleAuthR.verbose = 4),
  {
    googleCloudStorageR::gcs_auth("~/google-pins.json")
    googleCloudStorageR::gcs_get_object(
      object_name = "big-numbers/20221220T011808Z-b89e8/big-numbers.json",
      bucket = "pins-dev",
      saveToDisk = tempfile()
    )
  }
)
#> ℹ Downloading big-numbers/20221220T011808Z-b89e8/big-numbers.json
#> ✔ Saved big-numbers/20221220T011808Z-b89e8/big-numbers.json to /var/folders/hv/…
#> 
#> [1] TRUE
```

<sup>Created on 2022-12-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I believe other functions like `googleCloudStorageR::gcs_upload()` and `googleCloudStorageR::gcs_list_objects()` _do_ respect the option. Would you like me to open an issue? I will use `suppressMessages()` for that one function for now.

juliasilge:
I'm going to keep `pin_list()` turned off for now, because it looks like we don't have the same support for handling the flat namespace that we do in Azure and AWS. Since we don't have `pin_list()`, that means we don't have the ability to [make a manifest file](https://pins.rstudio.com/dev/reference/write_board_manifest.html). If we get requests for this, we can look into our options for implementing `pin_list()`.

juliasilge:
@MarkEdmondson1234 do you have the bandwidth to do a review of this PR and see if you have any suggestions for better use of the GCS API?

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+  rlang::check_installed("sodium")
+  path_to_encrypted_json <- fs::path_package("pins", "secret", "pins-gcs-testing.json")
+  raw <- readBin(path_to_encrypted_json, "raw", file.size(path_to_encrypted_json))
+  pw <- Sys.getenv("PINS_GCS_PASSWORD", "")
+  json <- sodium::data_decrypt(
+    bin = raw,
+    key = sodium::sha256(charToRaw(pw)),
+    nonce = secret_nonce()
+  )
+  googleCloudStorageR::gcs_auth(json_file = rawToChar(json))
```
juliasilge, line 85: I ended up taking the approach outlined in the [gargle vignette "Managing tokens securely"](https://gargle.r-lib.org/articles/managing-tokens-securely.html), and it seems to be working great!

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+  board_gcs("pins-dev",
+            cache = tempfile(),
+            ...
+  )
+}
+
+#' @export
+pin_list.pins_board_gcs <- function(board, ...) {
+  NA
+}
```
juliasilge, line 96: Like I discussed earlier in this PR, no `pin_list()` for the time being because we don't have good tooling for handling the flat namespace in a bucket. This is different from AWS and Azure. We can try coming back to this later if it is a high priority issue for folks.

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+
+#' @export
+pin_versions.pins_board_gcs <- function(board, name, ...) {
+  check_pin_exists(board, name)
+  resp <- googleCloudStorageR::gcs_list_objects(
+    bucket = board$bucket,
+    prefix = paste0(board$prefix, name)
+  )
+  paths <- fs::path_split(unique(fs::path_dir(resp$name)))
+  version_from_path(map_chr(paths, ~ .x[[length(.x)]]))
```
juliasilge, line 121: This is what I did to deal with the flat namespace situation; suggestions welcome for a nicer option!

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+    type = "text/yaml",
+    name = key,
+    ...
+  )
+}
+
+gcs_download <- function(board, key) {
+  path <- fs::path(board$cache, key)
+  if (!fs::file_exists(path)) {
+    suppressMessages(googleCloudStorageR::gcs_get_object(
```
juliasilge, line 239: This one function doesn't seem to respect `googleAuthR.verbose` so for now we'll use `suppressMessages()`.

R/board_gcs.R
MarkEdmondson1234: Nice solution for local testing

R/board_gcs.R
MarkEdmondson1234: I guess to enable it this would need to download a list of all objects, then parse through it locally for those ending with /.  Its doable, but could get slow if a very large (10k+) number of objects

R/board_gcs.R
MarkEdmondson1234: I guess its the use of the cli:: message bar - yes please open an issue as should be a quick fix

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+      object_name = paste0(board$prefix, key),
+      bucket = board$bucket,
+      saveToDisk = path
+    ))
+    fs::file_chmod(path, "u=r")
+  }
+  path
+}
+
+gcs_file_exists <- function(board, name) {
```
MarkEdmondson1234, line 249: I would do this via `gcs_get_object("obj_name", meta=TRUE)` and then `tryCatch(http_404 = FALSE)` - will be much quicker and direct only inspecting HTTP headers and not content.  Custom error types are supported in the HTTP responses to help this workflow. Listing objects could be slow for large amount of pins

MarkEdmondson1234: An example here for a pull request I did for targets a while ago https://github.com/MarkEdmondson1234/targets/blob/8204e4268553a2680f5d5ff3b0fc006b0c40d45a/R/utils_gcp.R#L7-22

R/board_gcs.R
MarkEdmondson1234: There is the concept of object versioning in GCS, does this match what pin_versions does? https://cloud.google.com/storage/docs/object-versioning

R/board_gcs.R
juliasilge: Just opened here: https://github.com/cloudyr/googleCloudStorageR/issues/170

R/board_gcs.R
juliasilge: No, one of the ideas of pins is to provide versioning, even if the data practitioner is not allowed to turn on GCS versioning because of, say, the bucket retention policy.

R/board_gcs.R
juliasilge: Ah, I see what you're saying! Hmmm... 

It's possible this function could have a better name, but it's not always looking for a specific object; sometimes it is looking for a directory-like thing. Given the information at hand when calling, for example, `pin_exists()`, I think we do need to list the objects in the bucket rather than get a specific object.

R/board_gcs.R
juliasilge: You can see how we handle that for [AWS](https://github.com/rstudio/pins-r/blob/b50eb57ae5879e28f85e42940f8f4944b49ff471/R/board_s3.R#L146-L156) and [Azure](https://github.com/rstudio/pins-r/blob/main/R/board_azure.R#L241-L250). Both of those rely on underlying functionality already available in paws.storage and AzureStor (ways to identify directory-like structure, common prefixes, etc). I don't think I want to move forward with writing that parsing code here in pins for now; maybe we can collaborate in the future on adding features like this to googleCloudStorageR if it is a high priority for users.

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+#'   `predefinedAcl` and `upload_type` through the dots of `pin_write()`.
+#' * `board_gcs()` is powered by the googleCloudStorageR package, which is a
+#'   suggested dependency of pins (not required for pins in general). If
+#'   you run into errors when deploying content to a server like
+#'   <https://shinyapps.io> or [Connect](https://posit.co/products/enterprise/connect/),
+#'   add `library(googleCloudStorageR)` to your app or document for [automatic
+#'   dependency discovery](https://support.posit.co/hc/en-us/articles/229998627-Why-does-my-app-work-locally-but-not-on-my-RStudio-Connect-server).
+#'
+#' @inheritParams new_board
+#' @param bucket Bucket name. You can only write to an existing bucket. Defaults
```
hadley, line 28: It seems weird to give this a default to me. Not sure if I can articulate why.

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+#' }
+board_gcs <- function(bucket = googleCloudStorageR::gcs_get_global_bucket(),
+                      prefix = NULL,
+                      versioned = TRUE,
+                      cache = NULL) {
+
+  check_installed("googleCloudStorageR")
+
+  # Check that have access to the bucket
+  invisible(googleCloudStorageR::gcs_get_bucket(bucket))
```
hadley, line 57:
```suggestion
  googleCloudStorageR::gcs_get_bucket(bucket)
```
?

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+
+  cache <- cache %||% board_cache_path(paste0("gcs-", bucket))
+  new_board_v1("pins_board_gcs",
```
hadley, line 60: Indenting needs a fix. (Do you know about https://github.com/lionel-/codegrip?)

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+}
+
+board_gcs_test <- function(...) {
+
+  skip_if_missing_envvars(
+    tests = "board_gcs()",
+    envvars = c("PINS_GCS_PASSWORD")
+  )
+
+  rlang::check_installed("sodium")
```
hadley, line 76: FWIW we generally just assume suggested packages are available for testing: https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-suggests-in-tests

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+  raw <- readBin(path_to_encrypted_json, "raw", file.size(path_to_encrypted_json))
+  pw <- Sys.getenv("PINS_GCS_PASSWORD", "")
+  json <- sodium::data_decrypt(
+    bin = raw,
+    key = sodium::sha256(charToRaw(pw)),
+    nonce = secret_nonce()
+  )
+  googleCloudStorageR::gcs_auth(json_file = rawToChar(json))
+
+  board_gcs("pins-dev",
```
hadley, line 87: Indenting

R/board_gcs.R
hadley: Why can't we use `gcs_list_objects()` plus a prefix? Because it returns all objects "recursively"?

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+
+# Helpers -----------------------------------------------------------------
+
+gcs_delete_dir <- function(board, dir) {
+  resp <- googleCloudStorageR::gcs_list_objects(
+    bucket = board$bucket,
+    prefix = paste0(board$prefix, dir, "/")
+  )
+
+  if (nrow(resp) == 0) {
```
hadley, line 213: I think this is probably not needed unless when `nrow()` is a 0, there's no `name` column.

R/board_gcs.R
```diff
@@ -0,0 +1,262 @@
...
+gcs_file_exists <- function(board, name) {
+  resp <- googleCloudStorageR::gcs_list_objects(
+    bucket = board$bucket,
+    prefix = paste0(board$prefix, name)
+  )
+  nrow(resp) > 0
+}
+
+## for decrypting JSON for service account:
+secret_nonce <- function() {
```
hadley, line 258: Move this up next to `board_gcs_test()`?

hadley (approved):
LGTM!

R/board_gcs.R
juliasilge: Yeah, none of our other boards uses this kind of default. I saw that googleCloudStorageR has this support for setting up a default bucket to use via environment variable, and I thought I would integrate it in case people are already doing this. If you feel strongly, I am fine with removing it.

R/board_gcs.R
juliasilge: Yes, that's right:

``` r
library(pins)
googleCloudStorageR::gcs_auth("~/google-pins.json")
b <- board_gcs("pins-testing")
googleCloudStorageR::gcs_list_objects(bucket = b$bucket)
#>                                                         name      size
#> 1        big-numbers/20230110T204627Z-b89e8/big-numbers.json  46 bytes
#> 2                big-numbers/20230110T204627Z-b89e8/data.txt 187 bytes
#> 3              small-numbers/20230110T204626Z-4d1eb/data.txt 189 bytes
#> 4     small-numbers/20230110T204626Z-4d1eb/small-numbers.rds  61 bytes
#> 5              small-numbers/20230110T204628Z-c3943/data.txt 191 bytes
#> 6    small-numbers/20230110T204628Z-c3943/small-numbers.json  23 bytes
#> 7            test-kcH7P3OJjX/20230110T205041Z-6e4eb/data.txt 192 bytes
#> 8 test-kcH7P3OJjX/20230110T205041Z-6e4eb/test-kcH7P3OJjX.rds  44 bytes
#>               updated
#> 1 2023-01-10 20:46:28
#> 2 2023-01-10 20:46:28
#> 3 2023-01-10 20:46:26
#> 4 2023-01-10 20:46:27
#> 5 2023-01-10 20:46:29
#> 6 2023-01-10 20:46:29
#> 7 2023-01-10 20:50:41
#> 8 2023-01-10 20:50:41
googleCloudStorageR::gcs_list_objects(bucket = b$bucket, prefix = "big-numbers")
#>                                                  name      size
#> 1 big-numbers/20230110T204627Z-b89e8/big-numbers.json  46 bytes
#> 2         big-numbers/20230110T204627Z-b89e8/data.txt 187 bytes
#>               updated
#> 1 2023-01-10 20:46:28
#> 2 2023-01-10 20:46:28
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

In the packages for AWS and Azure, there is existing support for identifying directory-like structures or to find "common prefixes" but googleCloudStorageR doesn't have that as of today.

R/board_gcs.R
juliasilge: Yep, that is the situation:

``` r
library(pins)
googleCloudStorageR::gcs_auth("~/google-pins.json")
b <- board_gcs("pins-testing")
resp <- googleCloudStorageR::gcs_list_objects(
  bucket = b$bucket,
  prefix = paste0(b$prefix, "cats-and-dogs", "/")
)
#> ℹ 2023-01-12 10:50:27 > No objects found
resp
#> data frame with 0 columns and 0 rows
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #697, opened 2023-01-11
Make tags more consistent in YAML

juliasilge:
From @machow in rstudio/pins-python#175:

> @juliasilge WDYT of us having R pins add tags metadata as either null or a list of strings? Currently, if you `pin_write(..., tags="some_tag")` it saves it as a string, rather than a list of strings.
> 
> AFAIK the result will be the same when reading meta into R (NULL or a vector of strings), so this is purely for convenience on the python side :) (which would be None, str, or list[str] right now)

With the change in this PR, now we will see the `tags` metadata represented the same in YAML whether there is one or more than one:

``` r
library(pins)
b <- board_temp()

b %>% pin_write(1:10, "nice-numbers", tags = "a-single-tag")
#> Guessing `type = 'rds'`
#> Creating new version '20230111T022854Z-4d1eb'
#> Writing to pin 'nice-numbers'
b %>% pin_write(100:10, "big-numbers", tags = c("a-single-tag", "another-tag"))
#> Guessing `type = 'rds'`
#> Creating new version '20230111T022854Z-02d96'
#> Writing to pin 'big-numbers'

cat(readr::read_lines(fs::path(
  b$path, 
  "nice-numbers", 
  pin_versions(b, "nice-numbers")$version, 
  "data.txt"
)), sep = "\n")
#> file: nice-numbers.rds
#> file_size: 61
#> pin_hash: 4d1eb93dc2188114
#> type: rds
#> title: 'nice-numbers: a pinned integer vector'
#> description: ~
#> tags:
#> - a-single-tag
#> created: 20230111T022854Z
#> api_version: 1.0

cat(readr::read_lines(fs::path(
  b$path, 
  "big-numbers", 
  pin_versions(b, "big-numbers")$version, 
  "data.txt"
)), sep = "\n")
#> file: big-numbers.rds
#> file_size: 183
#> pin_hash: 02d96ec26791911a
#> type: rds
#> title: 'big-numbers: a pinned integer vector'
#> description: ~
#> tags:
#> - a-single-tag
#> - another-tag
#> created: 20230111T022854Z
#> api_version: 1.0
```

<sup>Created on 2023-01-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

`NULL` values for tags also still work. 👍 

Related to #677


juliasilge:
@machow Want to make sure this is what you were looking for?

machow (approved):
Sorry for the wait -- I just tested it out and it LGTM! Thanks for making these changes!

PR #706, opened 2023-02-08
New `board_connect_url()`

### LINKED ISSUE #616: Provide a helper function to download pins straight from Connect content/vanity URLs
Right now, three pieces of information (perhaps four, depending on how you look at it) are required for reading a pin from RStudio Connect:

1. The Connect server URL
2. the full name of the pin ("username/pin_name" - You might count these separately. 🤷‍♀️)
3. Your API key

This increases friction for Connect users since they're more accustomed to sharing either content URLs or vanity URLs. More pieces of information to share means more chance for errors.

If your objective is to read a pin from Connect with the least amount of friction, the above is too much information.

If we provided a function that could download directly from a Connect content URL (either a standard content URL, which are very long or a vanity URL), we would reduce the number of pieces of required information to two.

1. The content/vanity URL
2. Your API key

Something like:

``` R
library(pins)
read_pin_from_connect("https://connect.example.com/my_penguin_pin", api_key = "xxxxxxxxxxxxx")
```

This would be a convenience function to load the latest version of the pin from Connect only. It would not offer the full power of working with a board

juliasilge:
Closes #616 

For testing, I'll need to write a pin or two with `board_connect()` and then set the vanity URL:
https://docs.posit.co/connect/api/#put-/v1/content/%7Bguid%7D/vanity


R/board_connect.R
```diff
@@ -726,12 +732,6 @@ connect_has_colorado <- function() {
   accounts <- rsconnect::accounts()
   any(c("colorado.rstudio.com", "colorado.posit.co") %in% accounts$server)
 }
-board_connect_colorado <- function(...) {
-  if (!connect_has_colorado()) {
-    testthat::skip("board_connect_colorado() only works with Posit's demo server")
-  }
-  board_connect(..., server = "colorado.posit.co", auth = "rsconnect", cache = fs::file_temp())
-}
```
juliasilge, line 734: This was in here twice. 🤷‍♀️

R/board_connect_url.R
```diff
@@ -0,0 +1,155 @@
...
+                              key = NULL,
+                              cache = NULL,
+                              use_cache_on_failure = is_interactive()) {
+
+  server <- rsc_server(auth, server, account, key)
+  cache <- cache %||% board_cache_path(paste0("connect-url-", hash(server$url)))
+
+  board <- new_board_v1(
+    "pins_board_connect_url",
+    vanity_urls = vanity_urls,
```
juliasilge, line 42: I am calling this `vanity_urls` for now until I figure out if we need the `url` element in the board still. Maybe we end up with just plain `urls`, in which case, I can reuse some of the `board_url()` methods.

juliasilge: Turns out I'm not going to use `urls` because I need the `url` element on the board to use `rsc_GET()` and similar. So let's stick with `vanity_urls` to distinguish it.

R/board_connect_url.R
```diff
@@ -0,0 +1,154 @@
...
+#' @description
+#' `board_connect_url()` lets you build up a board from individual
+#' [vanity urls](https://docs.posit.co/connect/user/content-settings/#custom-url).
+#' To use a Posit Connect board like this one or [board_connect()], you need to
+#' first authenticate. The easiest way to do so is by launching **Tools** -
+#' **Global Options** - **Publishing** - **Connect**, and follow the
+#' instructions.
+#'
+#' `board_connect_url()` is read only, and does not support either the use of a
+#' [manifest file][write_board_manifest()] or versioning.
```
juliasilge, line 12: We can't use the approach to handle versioning that `board_url()` uses (no manifest files on Connect) or that `board_connect()` uses (no looking up GUIDs), so no versioning on this board.

R/board_connect_url.R
```diff
@@ -0,0 +1,177 @@
...
+}
+
+#' @export
+pin_fetch.pins_board_connect_url <- function(board, name, version = NULL, ...) {
+  pin_fetch.pins_board_connect(board, name, version = NULL, ...)
+}
+
+#' @rdname board_deparse
+#' @export
+board_deparse.pins_board_connect_url <- function(board, ...) {
```
juliasilge, line 100: I did decide to create a `board_deparse` method for this board (`board_url()` does not have one FYI). This method is mostly used in vetiver and a board _without_ versioning is not the best fit for MLOps tasks, but I tend to think it's better to support the deparsing than not.

R/board_connect_url.R
```diff
@@ -0,0 +1,177 @@
...
+    "pins_board_connect_url",
+    vanity_urls = vanity_urls,
+    cache = cache,
+    url = server$url,
+    auth = server$auth,
+    use_cache_on_failure = use_cache_on_failure
+  )
+
+  version <- try_catch_rsc_version(board, cache, server)
+  pins_inform("Connecting to Posit Connect {version} at <{server$url}>")
```
juliasilge, line 50: I could put this inside of the new helper function `try_catch_rsc_version()`?

R/board_connect.R
```diff
@@ -724,13 +730,7 @@ board_connect_test <- function(...) {
...
 connect_has_colorado <- function() {
   accounts <- rsconnect::accounts()
-  any(c("colorado.rstudio.com", "colorado.posit.co") %in% accounts$server)
-}
-board_connect_colorado <- function(...) {
-  if (!connect_has_colorado()) {
-    testthat::skip("board_connect_colorado() only works with Posit's demo server")
-  }
-  board_connect(..., server = "colorado.posit.co", auth = "rsconnect", cache = fs::file_temp())
+  any(c("colorado.posit.co") %in% accounts$server)
```
hadley, line 733:
```suggestion
  "colorado.posit.co") %in% accounts$server
```

juliasilge:
Closed by #732 

I did end up re-using all that testing code I wrote in this PR so yay! 🎉

PR #715, opened 2023-02-15
Update vignette for how to customize metadata for specific use case

### LINKED ISSUE #456: pin_write/pin_upload generics
e.g.

```R
pin_write <- function(board, x,
                      name = NULL,
                      type = NULL,
                      desc = NULL,
                      metadata = NULL,
                      versioned = NULL,
                      ...) {
  ellipsis::check_dots_used()
  check_board(board, "pin_write()", "pin()")

  UseMethod("pin_write")
}

pin_write.pins_board <- function(board, x,
                      name = NULL,
                      type = NULL,
                      desc = NULL,
                      metadata = NULL,
                      versioned = NULL,
                      ...) {
  ...
}

```

Then `board_rsconnect()` can document its methods:

```R
pin_write.pins_board_rsconnect <- function(..., access_type = c("acl", "logged_in", "all")) {
  NextMethod()
}
```

Need to consider whether or not this additional layer is worth the effort.

juliasilge:
Closes #456 

This PR adds a new section to the ["Managing custom formats"](https://pins.rstudio.com/articles/managing-custom-formats.html) vignette about how to wrap `pin_write()` / `pin_read()` for consistent metadata.

### DIFF
--- a/vignettes/customize-pins-metadata.Rmd
+++ b/vignettes/customize-pins-metadata.Rmd
@@ -1,0 +1,87 @@
+---
+title: "Create consistent metadata for pins"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Create consistent metadata for pins}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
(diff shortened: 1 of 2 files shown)


juliasilge:
@machow Can you take a look at this vignette update about examples for wrapping pins functions?

machow:
Sorry for the long wait. I really like the new metadata example! I wonder if it would do well in its own article? It seems like a very big usecase, since it's the one vetiver uses, but felt a bit tucked away. I think part of it is that put all-together, it's a single point on this list, but a big point!:

- CSV files using readr or vroom
- Arrow files without using compression
- Parquet files
- Whole directories that are archived/zipped
- Objects with standardized metadata

PR #717, opened 2023-02-17
Follow up inspired by @kmasiello's comments

slodge:
Yet more followup .... https://github.com/rstudio/pins-r/issues/616

This enables (slow) url only code like:

```
devtools::load_all()
data <- pin_read("https://connect-uat.example.com/example_mtcars_pin/")
pin_meta("https://connect-uat.example.com/example_mtcars_pin/")
pin_versions("https://connect-uat.example.com/example_mtcars_pin/")
pin_write("https://connect-uat.example.com/example_mtcars_pin/", mtcars %>% head(10))
pin_versions("https://connect-uat.example.com/example_mtcars_pin/")
```

as well as (faster) code like:
```
devtools::load_all()
# might be better named as board_auto_single or board_single or similar?
vanity_board <- board_auto("https://connect-uat.example.com/example_mtcars_pin/") 
pin_read(vanity_board)
pin_meta(vanity_board)
pin_versions(vanity_board)
pin_write(vanity_board, mtcars %>% head(10))
```

or (more conventional) like:

```
devtools::load_all()

board <- board_connect(server="connect-uat.example.com")
pins <- pin_list(board)

pin_read(board, "https://connect-uat.example.com/example_mtcars_pin/")
pin_versions(board, "https://connect-uat.example.com/example_mtcars_pin/")
pin_meta(board, "https://connect-uat.example.com/example_mtcars_pin/")
pin_write(board, mtcars %>% head(4), "https://connect-uat.example.com/example_mtcars_pin/")
pin_versions(board, "https://connect-uat.example.com/example_mtcars_pin/")
pin_write(board, mtcars, "https://connect-uat.example.com/example_mtcars_pin/")
pin_versions(board, "https://connect-uat.example.com/example_mtcars_pin/")

pin_read(board, "https://connect-uat.example.com/pin_2/")
```

The code should be fairly easily to extend to e.g. `pin_version_delete` etc  .... but:
- it's slightly perverting the API...  
- it's also slow and fairly Connect specific at present... caching might help - or we could just accept slowness for people who want vanity access like this (as soon as Connect roadmap unwinds then perf will improve)
- it currently copes which Connect auto and envvar auth modes (makes some assumption) and falls back to board_url... but does not have any code for AWS, GCP, etc

Just here for discussion and for me to "learn things" - not as an actual PR proposal


NAMESPACE
```diff
@@ -93,15 +93,6 @@ S3method(pin_load,default)
 S3method(pin_load,files)
 S3method(pin_load,package)
 S3method(pin_load,table)
-S3method(pin_meta,pins_board_azure)
```
slodge, line 96: these changed to non-S3 methods because I haven't read enough of AdvancedR

juliasilge:
Thanks so much for your patience on this feature! 🙌 We landed on an option in #732 that we feel good about both adding now and maintaining/supporting in the long run:

``` r
library(pins)
board <- board_connect_url(c(
  numbers = "https://colorado.posit.co/rsc/some-nice-numbers/"    ## this is a private pin
))
board %>% pin_read("numbers")
#>  [1]  1  2  3  4  5  6  7  8  9 10
```

<sup>Created on 2023-04-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This _only_ works with environment variable authentication, and you can see [new documentation here](https://pins.rstudio.com/dev/reference/board_connect_url.html). You would install with `remotes::install_github("rstudio/pins-r")` to use this right now.

PR #719, opened 2023-02-22
S3 pagination implementation

mzorko:
Pagination implementation for S3 board of the `pin_list.pins_board_s3` & `pin_versions.pins_board_s3`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pins (development version)
 
+* `board_s3()` now uses pagination for listing and versioning (#719, @mzorko).
+
 # pins 1.1.0
 
 ## Breaking changes
--- a/R/board_s3.R
+++ b/R/board_s3.R
@@ -96,9 +96,9 @@ board_s3 <- function(
     region = NULL,
     endpoint = NULL,
     cache = NULL) {
-  
+
   check_installed("paws.storage")
-  
+
   config <- compact(list(
     credentials = compact(list(
       creds = compact(list(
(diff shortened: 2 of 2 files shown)


juliasilge:
Thank you so much for this contribution @mzorko! Is there a way to update this PR (or make a new one) to not include the whitespace changes?

mzorko:
Hi @juliasilge. Sure, sorry for that - there is a new commit included.

juliasilge:
@mzorko This looks great to me. Do you think it's possible to write a test for this that would be straightforward and not too slow? Or would it require writing just a zillion pins (or versions)? You can check out the suites of tests we run across the boards ([like this](https://github.com/rstudio/pins-r/blob/main/tests/testthat/test-board_s3.R)) but this would be an S3-specific test, I think.

mzorko:
@juliasilge Hmm...I did some local tests yesterday and it took more than 1 hour to create 1,000+ pins in S3 with iris data (AWS slows down significantly after 200 API calls). Maybe it would make sense do do it once to create S3 pins example with more than 1,000 keys and to use that for testing. Will check current state of tests to see if we have any public bucket available for testing and to think how to implement this.

juliasilge:
Hmmmm, we do use a [specific bucket](https://github.com/rstudio/pins-r/blob/main/R/board_s3.R#L130) for testing but our testing strategy now writes/reads and then tears it all down too; we don't have anything that persists across testing sessions. Let's say that writing a test for this will be impractical currently, and go for it as is.

juliasilge (approved):
Thank you again for this contribution @mzorko! 🚀

Issue #606, opened 2022-03-29
Support pin_search on RStudio Connect when user is an admin


machow:
See https://github.com/machow/pins-python/issues/27#issuecomment-1075633988

Currently, admin users on RStudio Connect can show a list of all pin content, including content they are not able to access. However, this leads to this bad situation:

* admin user runs pin_search, it
* searches all pins, including those they can't access
* for matching pins, it attempts to access metadata

In order to resolve, we could catch failures to fetch metadata and have them return something with less information (e.g. pin name).

See:

* [where RSC calls multi_meta](https://github.com/rstudio/pins/blob/main/R/board_rsconnect.R#L325), which calls pin_meta, which attempts to fetch data.

kmasiello:
Also posting a reminder that `pin_list()` and `pin_exists()` may be suitable alternatives for users running into this issue, depending on what the use case is for running `pin_search()`.

PR #722, opened 2023-02-24
Actually import purrr

juliasilge:
This PR sets up pins to use purrr directly, instead of using the older `compat-purrr.R` file (which is very out of date, doesn't have adverbs like `possibly()`, etc).


juliasilge:
We were at 20 imported packages, so the first step I took here was to move several packages to Suggests that are only used in one function or only in legacy functions.

juliasilge:
Not sure if you want to take a look at this or not @hadley -- let me know if not!

tests/testthat/test-legacy_local.R
```diff
@@ -1,3 +1,5 @@
+skip_if_not_installed("filelock")
```
hadley, line 1: FWIW our general policy is not to add skips: https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-suggests-in-tests

R/board_connect_bundle.R
```diff
@@ -52,6 +52,7 @@ rsc_bundle_preview_create <- function(board, name, x, metadata, path) {
 }
 
 rsc_bundle_preview_index <- function(board, name, x, metadata) {
+  check_installed("whisker")
```
hadley, line 55: I think it's probably worth leaving whisker in imports, since if you're using Connect, you'll hit this path?

hadley (approved):
I don't think this specific PR uses any purrr features, but I presume you have some plans further down the line.

tests/testthat/test-legacy_local.R
juliasilge: Oh, interesting. I've run into problems with my tests using a suggested package failing on CRAN when the package gets archived. This kind of problem is partly why I/we/tidymodels starting doing a ["hard" `R CMD check`](https://github.com/rstudio/pins-r/blob/main/.github/workflows/R-CMD-check-hard.yaml), to avoid getting in that situation. Personally I've found it worthwhile for packages like vetiver, pins, broom, etc that provide methods for a lot of different packages. FWIW that GH action requires all the use of suggested packages in tests to be protected with skips.

juliasilge:
> I don't think this specific PR uses any purrr features, but I presume you have some plans further down the line.

Yep, that's right. I've been running into some annoyances because of not having real purrr and this particular PR was finally motivated by me trying to fix #606 using `possibly()`.

PR #723, opened 2023-02-25
Make `multi_meta()` robust to failures in `pin_meta()`

### LINKED ISSUE #606: Support pin_search on RStudio Connect when user is an admin

juliasilge:
Closes #606 

On Connect, admins can list _all_ content but cannot call `pin_meta()` on content they don't have access to. This PR wraps the call to `pin_meta()` in `possibly()` and returns `list()` when it fails, much like this for when `names` is empty:

https://github.com/rstudio/pins-r/blob/31573aa947b61768e397e31813dabe811ad8bc7a/R/pin-meta.R#L52-L60

There is a pin on our demo board by @kmasiello that I don't have access to, `"katie.masiello/LETTERS"`. With the current proposal, I would see this:

``` r
library(pins)
b <- board_connect()
#> Connecting to Posit Connect 2023.01.0 at <https://colorado.posit.co/rsc>

## existing behavior for empty `names`:
pins:::multi_meta(b, NULL)
#> # A tibble: 0 × 6
#> # … with 6 variables: name <chr>, type <chr>, title <chr>, created <dttm>,
#> #   file_size <fs::bytes>, meta <list>

# proposed new behavior when `pin_meta` fails:
pins:::multi_meta(b, c("julia.silge/cat-tracking-uk", "katie.masiello/LETTERS"))
#> # A tibble: 2 × 6
#>   name                        type  title created             file_…¹ meta      
#>   <chr>                       <chr> <chr> <dttm>              <fs::b> <list>    
#> 1 julia.silge/cat-tracking-uk arrow juli… 2023-02-15 12:14:40    689K <pins_met>
#> 2 katie.masiello/LETTERS      <NA>  <NA>  NA                        0 <list [0]>
#> # … with abbreviated variable name ¹​file_size
```

<sup>Created on 2023-02-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I like all of that pretty well, except for the zero file size. Better as `NA`? Or is this OK?

@machow can you confirm for me what the behavior is in Python in this situation?


juliasilge:
We could change how `file_size` is computed to `fs::as_fs_bytes(map_dbl(meta, ~ sum(.x$file_size %||% NA_real_)))` and get an `NA` there. Is it possible to end up with, say, one `NULL` value and one real value there in `file_size` for pins with more than one file? Because then we'd `sum()` that to `NA`. Maybe an OK outcome in that situation?

machow:
> @machow can you confirm for me what the behavior is in Python in this situation?

It looks like python pins uses `None` (roughly `NA`) for file_size when the admin can't access metadata:

```python
            name  type title created file_size                                               meta
0  susan/some_df  None  None    None      None  MetaRaw(file=None, type=None, name='susan/some...
```

juliasilge:
Thanks @machow! I updated what this returns now to make it more consistent with the Python implementation:

``` r
library(pins)
b <- board_connect()
#> Connecting to Posit Connect 2023.01.0 at <https://colorado.posit.co/rsc>

# new behavior when `pin_meta` fails:
pins:::multi_meta(b, c("julia.silge/cat-tracking-uk", "katie.masiello/LETTERS"))
#> # A tibble: 2 × 6
#>   name                        type  title created             file_…¹ meta      
#>   <chr>                       <chr> <chr> <dttm>              <fs::b> <list>    
#> 1 julia.silge/cat-tracking-uk arrow juli… 2023-02-15 12:14:40    689K <pins_met>
#> 2 katie.masiello/LETTERS      <NA>  <NA>  NA                       NA <pins_met>
#> # … with abbreviated variable name ¹​file_size
```

<sup>Created on 2023-02-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

juliasilge:
@kmasiello as an admin on our own demo server, would you be up for testing out this solution for `pin_search()` to see if you have any feedback? You can install via `devtools::install_github("rstudio/pins-r@multi-meta-possibly")` and then I would expect `pin_search()` to work for you (although it still won't be very speedy).

R/pin-meta.R
```diff
@@ -66,7 +66,7 @@ multi_meta <- function(board, names) {
       type = map_chr(meta, ~ .x$type %||% NA_character_),
       title = map_chr(meta, ~ .x$title %||% NA_character_),
       created = .POSIXct(map_dbl(meta, ~ .x$created %||% NA_real_)),
-      file_size = fs::as_fs_bytes(map_dbl(meta, ~ sum(.x$file_size) %||% NA_real_)),
+      file_size = fs::as_fs_bytes(map_dbl(meta, ~ sum(.x$file_size %||% NA_real_))),
```
juliasilge, line 69: So the `file_size` is returned as `NA` and not 0 when we don't have it

kmasiello:
@juliasilge, tested in local IDE and Workbench. Works without errors now 🎉 , though yes, it is a slow process.  On Workbench it took 4-9 minutes to run the search with no arguments specified, returning 846 records.  On local it was as much as 15 min.  Performance is much better if I put any refinement into the `search=` argument, speeding things up to less than 30 seconds.

juliasilge:
Oooof @kmasiello that is so rough 😩 but is more about the Connect API than code here in pins. At least we have a fix for the bad (buggy?) behavior that returns an error now.

juliasilge:
@hadley let me know if you don't want to review this! 🙌

kmasiello:
> 

Yes, it is a huge improvement. I kind of know what I'm stepping into when I perform an unrefined query though. If only we could support searching metadata.

PR #724, opened 2023-03-01
Remove uses of `is_testing()`

### LINKED ISSUE #640: Use of `is_testing()` is a problem for folks who program against pins functions

juliasilge:
Closes #640

### DIFF
--- a/R/board_azure.R
+++ b/R/board_azure.R
@@ -254,11 +254,11 @@ azure_dir_exists <- function(board, path) {
   AzureStor::storage_dir_exists(board$container, dir)
 }
 
-local_azure_progress <- function(progress = !is_testing(), env = parent.frame()) {
+local_azure_progress <- function(progress = is_interactive(), env = parent.frame()) {
   withr::local_options(list(azure_storage_progress_bar = progress), .local_envir = env)
 }
 
-azure_download <- function(board, keys, progress = !is_testing()) {
+azure_download <- function(board, keys, progress = is_interactive()) {
   local_azure_progress(progress)
 
   paths <- fs::path(board$cache, keys)
--- a/R/pin_versions.R
+++ b/R/pin_versions.R
@@ -146,11 +146,7 @@ versions_template <- function(version = character()) {
 }
 
 version_name <- function(metadata) {
-  if (is_testing()) {
-    paste0("20120304T050607Z-", substr(metadata$pin_hash, 1, 5))
-  } else {
-    paste0(metadata$created, "-", substr(metadata$pin_hash, 1, 5))
-  }
+  paste0(metadata$created, "-", substr(metadata$pin_hash, 1, 5))
 }
 
 version_from_path <- function(x) {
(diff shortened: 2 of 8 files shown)


juliasilge:
I am having trouble with the tests similar to r-lib/mockery#57 and [it looks like Gabor says](https://community.rstudio.com/t/mockery-stub-is-creating-mocking-functions-that-persist-into-other-tests/129259/7) that the `depth` argument doesn't really work. I'm having trouble currently writing a wrapper function that can successfully be stubbed out.

juliasilge:
The new testthat was just submitted so I think I'll 

- wait until that gets accepted
- remove all the installation of devel testthat
- have someone review at that point

juliasilge:
It's taking a while for the new testthat to get on CRAN so let's do this instead:

- review now and merge
- later remove all the installation of devel testthat (before the next release)

R/pin-read-write.R
```diff
@@ -144,21 +144,21 @@ object_write <- function(x, path, type = "rds") {
...
-    saveRDS(x, path, version = 2)
-  } else {
-    # compression algorithm changed in 4.1
-    saveRDS(x, path, version = 2, compress = FALSE)
+  saveRDS(x, path, version = 2)
+  invisible(path)
+}
 
-    old <- readBin(path, "raw", fs::file_size(path))
+write_rds_test <- function(x, path) {
```
juliasilge, line 151: This is only used in one test, but the function is called directly so I can't seem to mock it. If I do this:

```r
local_mocked_bindings(write_rds = write_rds_test)
```
it does not mock the results when I call `write_rds()`. Is that right?

tests/testthat/test-meta.R
```diff
@@ -5,7 +5,7 @@ test_that("non-existent file returns default data", {
 test_that("standard metadata is useful", {
   df <- data.frame(x = 1:10)
   path <- withr::local_tempfile()
-  write_rds(df, path)
+  write_rds_test(df, path)
```
juliasilge, line 8: Here is where I am calling `write_rds_test()`, that function used only here because I can't seem to mock `write_rds()` if I call it directly.

Issue #600, opened 2022-03-09
Feature request: board_github() for updated API


justinmillar:
It would be awesome if a future release could get an updated version of `register_board_github()` for the new API. In our use case, we've been creating data packages that use `pins` to provide up-to-date data without requiring package re-installation. The Github workflow is nice because we can keep the package code and datasets together within a single repo. Unfortunately `board_url()` doesn't quite get there since is it read-only, so we're still using the legacy API and are unable to use some of the new features (such as arrow support).

Thanks, and thank you for this excellent package!

juliasilge:
Thanks for your patience on this issue! Would it work for your use case to:

- write to a board using `board_folder()` (where the folder is something in the git repo where you want to keep this data)
- use the [new `write_board_manifest()` function](https://pins.rstudio.com/dev/reference/write_board_manifest.html) to specify what pins are there
- push these files
- have folks read via `board_url()` using the changes in #681 

Reading would look like this (check [out the folder here](https://github.com/rstudio/pins-r/tree/main/tests/testthat/pin-board)):

``` r
library(pins)
folder_on_gh <- "https://raw.githubusercontent.com/rstudio/pins-r/main/tests/testthat/pin-board/"
b <- board_url(folder_on_gh)

b %>% pin_list()
#> [1] "x" "y"
b %>% pin_versions("y")
#> # A tibble: 2 × 3
#>   version                created             hash 
#>   <chr>                  <dttm>              <chr>
#> 1 20221215T180357Z-9ae7a 2022-12-15 11:03:57 9ae7a
#> 2 20221215T180400Z-b81d5 2022-12-15 11:04:00 b81d5
```

<sup>Created on 2022-12-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

justinmillar:
Thanks for picking this up!

Yes, I think this workflow would work great for our use case. Currently we are using pins inside of an internal R package, which has a wrapper function around `pin_get()` to load in data. The package is stored in a private Github repo, and the pinned datasets are stored in a separate branch within the package repo. 

Would it still be possible to write using `board_folder()` to a private Github repo (assuming proper credentials)?

juliasilge:
Ah gotcha, an internal R package in a private repo. 👍  

You can use `board_folder()` to write to the private repo (because then the credentials are handled with `git push` and such) but you can't read via `board_url()` from the private repo unless you specify the exact path to the YAML and put the token on the URL itself, like this:

``` r
[...]

PR #732, opened 2023-04-10
New `add_headers` argument to `board_url()`

juliasilge:
This PR adds a new `add_headers()` argument to `board_url()`, mostly for authentication.

Addresses #600 for private GH repos:

``` r
library(pins)
folder_on_gh <- "https://raw.githubusercontent.com/juliasilge/pkgpins/main/pkgdown/assets/pins-board/_pins.yaml"
my_headers <- httr::add_headers(Authorization = paste("token", "github_pat_XXXX"))
b <- board_url(folder_on_gh, add_headers = my_headers)
b %>% pin_list()
#> [1] "mtcars_metric"
```

<sup>Created on 2023-04-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Addresses #616 for Posit Connect vanity URLs that are not set to "Anyone - no login required":

``` r
library(pins)
board <- board_url(
  c(my_vanity_url_pin = "https://colorado.posit.co/rsc/some-nice-numbers/"),
  add_headers = httr::add_headers(Authorization = paste("Key", Sys.getenv("CONNECT_API_KEY")))
)
board %>% pin_read("my_vanity_url_pin")
#>  [1]  1  2  3  4  5  6  7  8  9 10
```

<sup>Created on 2023-04-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


juliasilge:
Upon further consideration, the solution in #706 doesn't really get us enough of what folks are wanting, and is quite a big change. This more simple change results in more consistent versioning options (does require a manifest for *reading* different versions) and supports more use cases.

Ideally in the long term, we will have better options on Connect like service accounts that help with some of the same user needs. I think this PR is a good option to implement now, because it fits in with several use cases and is not a burdensome change.

tests/testthat/test-board_url.R
```diff
@@ -176,7 +176,7 @@ test_that("use cache with Last-Modified header", {
   url <- "https://httpbin.org/cache"
 
   expect_condition(
-    http_download(url, path, "test"),
+    http_download(url, path, "test", add_headers = httr::add_headers(x = 1)),
```
juliasilge, line 179: This is not extensive testing of the new argument, but I don't see a good way forward to really test authentication. Maybe in the Connect tests? Is this worth it?

juliasilge:
Partly to get some more thorough testing of whether this all works (but also because folks continue to ask), I ended up adding back in a `board_connect_url()` function which is a _very thin_ wrapper around `board_url()` with the new headers arg. It only authenticates to Connect via environment variable, which SE folks say is good because it is the only method that will work went content gets deployed to Connect.

We talked about whether it is possible to catch these kind of errors and provide a more helpful error message:

``` r
library(pins)

b2 <- board_url(
  c(non_public_numbers = "https://colorado.posit.co/rsc/some-nice-numbers/"),
)
b2 %>% pin_read("non_public_numbers")
#> Error in http_download(url = paste0(url, "data.txt"), path_dir = cache_dir, : Not Found (HTTP 404).
```

<sup>Created on 2023-04-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I don't know that we can for now, because Connect is returning a 404 where the real problem is lack of authentication (403).

tests/testthat/test-board_connect_url.R
```diff
@@ -0,0 +1,42 @@
+skip_if_not_installed("rsconnect")
```
juliasilge, line 1: This is needed because the test boards use `connect_has_colorado()`, not because `board_connect_url()` uses the rsconnect package (it does not).

R/board_url.R
```diff
@@ -304,11 +329,12 @@ http_download <- function(url, path_dir, path_file, ...,
     }
 
     headers <- httr::add_headers(
+      add_headers$headers,
```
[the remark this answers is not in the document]
juliasilge: This is because as it is here, the `add_headers` object was created with `httr::add_headers()` and you can't put those together:

``` r
h <- httr::add_headers(a = 1, b = 2)
h
#> <request>
#> Headers:
#> * a: 1
#> * b: 2

httr::add_headers(h, c = 3)
#> Error in f(init, x[[i]]): is.request(y) is not TRUE
httr::add_headers(h$headers, c = 3)
#> <request>
#> Headers:
#> * a: 1
#> * b: 2
#> * c: 3
```

<sup>Created on 2023-04-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I see what you're saying about having the argument use a named character instead and I'll check that out.

Issue #589, opened 2022-01-14
Feature Request: Re-enabling data versioning by hash only


afshinmhCLG:
**Desired behavior:**  
Data version changes when data changes. For example, repeated execution of `pin_write(x, board = myboard, version = T)`  does not generate different data versions as long as x is the same.

**Current behavior:**  
Data is re-written under a  new version each time `pin_write` is executed even when content, name, description, title, etc. are all the same! 

I realize hash of the data is stored so one can identify duplicates, but avoiding data duplication in the first place was at the core of the value `pins` offered teams. 

**Feature request:** 
Given that the time stamp was added to the version signature in version 1.0.1, I am sure it has a use-case, but would it be possible to enable an option where the signature is the hash of the data or even better a hash encompassing data, and other user defined metadata (e.g. description and other tags)?

robertsehlke:
+1

I'm just testing out pins, which so far has been very slick with s3 backend, and went looking in the issues if someone else had already asked about this.

For my use case, even just an optional duplicate detection built into `pin_write()` would serve the purpose. I.e. if the user chooses: calculate the hash, query existing versions, and only create a new version if it is not a duplicate.

This would be a simple QoL option that does not affect current structures.

machow:
It seems like the big challenge here for pins is that version name is made of two parts (`"{timestamp}-{hash}"`), and has two jobs:

* timestamp: allow for quick sorting of pin versions (e.g. to get the latest pin)
* hash: allow for quick checking whether contents are identical

It seems like versioning by hash only would require a new strategy for tracking timestamp info. For example, on `pin_write()` we check for a duplicate, and instead a filesystem `mv` to rename the duplicate version to the updated timestamp.

E.g. 

* a write of data with the hash `abc` on `2022-02-02`
* verifies a previous version has a matching hash `abc`
* moves `{old_timestamp}-abc` to `{new_timestamp}-abc`.

It seems like it might require fairly substantial changes, though....!

robertsehlke:
Agreed - a simpler way could be to let the user choose not to upload/update at all if there already is a version with the same hash (also keeps external references stable).

iainmwallace:
Agreed, this capability would be very useful.

perhaps there could be a helper function such as pin_unique_write that only writes if the dataset is unique.

machow:
Thanks for all the confirmation that this would be useful, and added context! I haven't seen the original pins in action, so it's helpful to hear that this feels like an original feature that got taken out.

[...]

Issue #592, opened 2022-01-25
`pin_write` should return the `version` as well as the `name`


sellorm:
In the the current version of pins (1.0.1) `pin_write()` returns the fully qualified name of the new pin, invisibly.

``` r
library(pins)
board = board_folder("~/tmp/pins")
pinout <- pin_write(board, 
                    ggplot2::diamonds, 
                    name = "diamonds", 
                    type = "csv")
#> Replacing version '20220125T122254Z-c9e9b' with '20220125T122531Z-c9e9b'
#> Writing to pin 'diamonds'
str(pinout)
#>  chr "diamonds"
```

There are a number of additional workflows that would be enabled by also returning the version. In the example above, `20220125T122531Z-c9e9b`.

This is useful, since both the "fully qualified name" _and_ the version are required to obtain a specific version of a data set rather than just the latest version.

For example, users may want to record the versions of specific pins in another system and currently, this is not a straightforward task.

Perhaps `pin_write()`, could return a list that includes the fully qualified domain name as well as the version.

afshinmhCLG:
Returning version would be a very useful feature! 
I rely on knowing the version prior to pinning so much that I have a wrapper/hack where I always pin into a local empty/temp location just to get the version and then pin to remote knowing what the version will be.

juliasilge:
We could look into whether it's possible to return the whole metadata object invisibly for `pin_write()`.

juliasilge:
It turns out the `meta` object in this function doesn't contain the name _or_ version (in fact, no metadata objects contain the name) so that's not really going to help us much here. 🙈 Instead, let's try returning a named list with `name` and `version`.

juliasilge:
In #735 we are looking at an argument to check the hash of the pin contents before pinning, so another possible component to consider returning is whether pinning was successful.

aluxh:
Just wanted to share my use case.

Returning name and version will be helpful for building the COPY INTO sql statement to ingest data into databases. Currently, I'm using `pin_versions(...) |> tail(1) |> pull(version)`

PR #735, opened 2023-04-26
New `force_identical_write` arg

juliasilge:
Addresses #589 

This does not solve all our hashing issues (for example, the `pin_hash` still only looks at the pin _contents_ and not the metadata) but it is a small step toward a better hashing situation.

After talking with users, I think that an opt-in new argument is the best way forward (rather than, say, a new function or new default behavior). This new `check_hash` arg comes after the dots and defaults to `FALSE`. It does require another read of `pin_meta()` to get the old hash.

``` r
library(pins)
b <- board_temp(versioned = TRUE)
b %>% pin_write(1:10, "nice-numbers", check_hash = TRUE)
#> Guessing `type = 'rds'`
#> Creating new version '20230426T224520Z-7f884'
#> Writing to pin 'nice-numbers'

## to change timestamp:
Sys.sleep(2)

b %>% pin_write(1:10, "nice-numbers", check_hash = TRUE)
#> Guessing `type = 'rds'`
#> Warning: The hash of pin "nice-numbers" has not changed and will not be stored.
```

<sup>Created on 2023-04-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

With this PR as is now, `pin_write()` will still return the name invisibly but does not store the pin and generates a warning. Thoughts?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,7 @@
 
 * Fixed bug in `cache_prune()` to correctly find caches for `board_url()` (#734).
 
-* Added new `check_hash` argument when writing a pin to check whether the pin 
+* `pin_write()` gains a `check_hash` argument to check whether the pin 
   contents are identical to the last version (#735).
 
 # pins 1.1.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,7 @@
 
 * Fixed bug in `cache_prune()` to correctly find caches for `board_url()` (#734).
 
-* `pin_write()` gains a `check_hash` argument to check whether the pin 
+* `pin_write()` gains a `compare_hash` argument to check whether the pin 
   contents are identical to the last version (#735).
 
 # pins 1.1.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # pins (development version)
 
+## Breaking changes
+
+* `pin_write()` no longer writes identical pin contents by default, and gains a
+  `force_identical_write` argument for writing even when the pin contents are 
+  identical to the last version (#735).
+
+## Other improvements
+
 * The `print` method for boards no longer calls `pin_list()` internally (#718).
 
 * `board_s3()` now uses pagination for listing and versioning (#719, @mzorko).
@@ -15,9 +23,6 @@
   well as new board for Connect vanity URLs `board_connect_url()` (#732).  
 
 * Fixed bug in `cache_prune()` to correctly find caches for `board_url()` (#734).
-
-* `pin_write()` gains a `compare_hash` argument to check whether the pin 
-  contents are identical to the last version (#735).
 
 # pins 1.1.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,9 @@
 
 * Fixed bug in `cache_prune()` to correctly find caches for `board_url()` (#734).
 
+* Added new `check_hash` argument when writing a pin to check whether the pin 
+  contents are identical to the last version (#735).
+
 # pins 1.1.0
 
 ## Breaking changes
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -119,9 +119,10 @@ pin_write <- function(board, x,
   if (check_hash) {
     old_hash <- possibly_pin_meta(board, name)$pin_hash
     if (!is.null(old_hash) && old_hash == meta$pin_hash) {
-      cli::cli_warn(
-        "The hash of pin {.val {name}} has not changed and will not be stored."
-      )
+      cli::cli_warn(c(
+        "!" = "The hash of pin {.val {name}} has not changed.",
+        "*" = "Your pin will not be stored."
+      ))
       return(invisible(name))
     }
   }
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -65,8 +65,8 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #' @param tags A character vector of tags for the pin; most important for
 #'   discoverability on shared boards.
 #' @param check_hash Check whether the pin contents are identical to the last
-#'   version (using a hash), and then **do not store** the pin again. Defaults
-#'   to `FALSE`.
+#'   version if one exists (using the hash), and then **do not store** the pin
+#'   again. Defaults to `FALSE`.
 #' @rdname pin_read
 #' @export
 pin_write <- function(board, x,
@@ -116,8 +116,7 @@ pin_write <- function(board, x,
   meta$user <- metadata
 
   if (check_hash) {
-    old_hash <- possibly_pin_meta(board, name)
-    old_hash <- old_hash$pin_hash
+    old_hash <- possibly_pin_meta(board, name)$pin_hash
     if (!is.null(old_hash) && old_hash == meta$pin_hash) {
       cli::cli_warn(
         "The hash of pin {.val {name}} has not changed and will not be stored."
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -66,7 +66,8 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   discoverability on shared boards.
 #' @param check_hash Check whether the pin contents are identical to the last
 #'   version if one exists (using the hash), and then **do not store** the pin
-#'   again. Defaults to `FALSE`.
+#'   again. This argument does not check the pin metadata, only the pin
+#'   contents. Defaults to `FALSE`.
 #' @rdname pin_read
 #' @export
 pin_write <- function(board, x,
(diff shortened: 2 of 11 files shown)


kmasiello:
Tested and works as expected. 
If I'm a user, I can work with approach well, though I think the documentation should be explicit in noting that the hash is not affected by changed metadata. I like that it's an opt-in argument, which means I will be reading the docs to understand what I'm getting into.
There might be situations where the user would want the status (pinned or didn't pin) to be returned in addition to the pin name. Every use case I envision though, I can noodle through how I'd determine the pinning status by other means. I think it would be wise to keep our ears open as to whether this is a need from the community that we'd implement in the future.

juliasilge:
Thank you so much @kmasiello! 🙌 

> There might be situations where the user would want the status (pinned or didn't pin) to be returned in addition to the pin name.

This is a good point. Right now, `pin_write()` only returns the `name` (like `"julia.silge/nice-numbers"`) but we have been discussing in #592 how to better organize what should be returned. I'll add this info to that issue and we can see what would work best.

R/pin-read-write.R
```diff
@@ -111,6 +116,16 @@ pin_write <- function(board, x,
   )
   meta$user <- metadata
 
+  if (check_hash) {
+    old_hash <- possibly_pin_meta(board, name)$pin_hash
+    if (!is.null(old_hash) && old_hash == meta$pin_hash) {
+      cli::cli_warn(
+        "The hash of pin {.val {name}} has not changed and will not be stored."
+      )
+      return(invisible(name))
```
juliasilge, line 125: We could return something different here, like `NULL` maybe.

juliasilge:
Refined the warning message a smidge:

``` r
library(pins)
b <- board_connect()
#> Connecting to Posit Connect 2023.03.0 at <https://colorado.posit.co/rsc>
b %>% pin_write(1:10, "julia.silge/really-nice-numbers", check_hash = TRUE)
#> Guessing `type = 'rds'`
#> Writing to pin 'julia.silge/really-nice-numbers'
b %>% pin_write(1:10, "julia.silge/really-nice-numbers", check_hash = TRUE)
#> Guessing `type = 'rds'`
#> Warning: ! The hash of pin "julia.silge/really-nice-numbers" has not changed.
#> • Your pin will not be stored.
b %>% pin_write(1:11, "julia.silge/really-nice-numbers", check_hash = TRUE)
#> Guessing `type = 'rds'`
#> Writing to pin 'julia.silge/really-nice-numbers'
```

<sup>Created on 2023-05-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kmasiello:
> Do we think that now in this PR `pin_write()` should return something different than the name if it does not update the pin? Or should we wait for broader changes to what is returned?

My gut says this should be handled with the broader considerations of what we return. I don't want to see a bespoke solution here that isn't compatible with future work.

kmasiello:
Additional suggestion and comments after reviewing with users - 
`check_hash` may be better named as `check_data_hash` to emphasize that metadata is not checked.

Checking metadata for changes is an outstanding ask. A sample scenario is using `check_hash = TRUE` but wanting to update the metadata to reflect a "current as of" date. You would want to be able to indicate the pin is still fresh, even if the data hasn't changed.

juliasilge:
> A sample scenario is using `check_hash = TRUE` but wanting to update the metadata to reflect a "current as of" date.

Can I make sure I am understanding this? It sounds like users are expressing a desire to:

- not fill up their disk space with copies of the same file, and _at the same time_
- have the metadata reflect the pin's freshness, like the last time a write was attempted (info more than just "latest")

This would require a more extensive change than in this PR, will be pretty tough in the near term, and likely would require special handling of different backends (S3 vs. Connect vs. folders vs. ...). As of now, updating the metadata but not the pin contents isn't something straightforward to implement. We could consider how to approach this problem in the future.

I tend to think we should stick with `check_hash` as the argument name since we will likely think through checking portions of the metadata as well (title, description, tags, and user metadata could all currently change without changing the pin hash).

I tend to think we should still go with the change as scoped here because it lets us offer a solution to the storing-many-copies problem. It does mean folks have to choose between having the metadata reflect the last time the pin contents _really_ changed and the last time a write was attempted; this still seems better than the situation as it is now.

kmasiello:
Agreed. As currently scoped, this helps the main use case tremendously. 
If I'm very keen to only store the changed data frame but have a record of the last write attempt, I can bake this logging info into my upstream ETL process. 
Similar to the other comment about returning the "wrote / didn't write" status, it will be good to keep ears open to use cases around metadata changes.

R/pin-read-write.R
```diff
@@ -111,6 +116,17 @@ pin_write <- function(board, x,
   )
   meta$user <- metadata
 
+  if (check_hash) {
+    old_hash <- possibly_pin_meta(board, name)$pin_hash
```
hadley, line 120: Why do we think this might fail?

R/pin-read-write.R
```diff
@@ -64,6 +64,10 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   use the default for `board`
 #' @param tags A character vector of tags for the pin; most important for
 #'   discoverability on shared boards.
+#' @param check_hash Check whether the pin contents are identical to the last
```
hadley, line 67: Should we consider defaulting this to `TRUE`?

NEWS.md
```diff
@@ -16,6 +16,9 @@
 
 * Fixed bug in `cache_prune()` to correctly find caches for `board_url()` (#734).
 
+* Added new `check_hash` argument when writing a pin to check whether the pin
```
hadley, line 19:
```suggestion
* `pin_write()` gains a `check_hash` argument to check whether the pin 
```

hadley (approved):
This seems so simple — I love it 😄

I also refreshed my memory of `pin_hash()`, and I'm confident it's ok to be used in this way. There is some chance that you might get the same hash value with different data values, but it's [extremely small](https://stackoverflow.com/questions/45788721/where-can-i-find-xxhash64-and-md5-collision-probability-statistics). (Or maybe [this is better](https://preshing.com/20110504/hash-collision-probabilities/) if you hashed a dataset every day for two years, you're more likely to get hit by a meteor than get a hash collision.)

R/pin-read-write.R
juliasilge: This is to let someone use `check_hash = TRUE` when they write a pin for the first time, before there is any metadata. `old_hash` returns as `NULL` here when there isn't any metadata.

R/pin-read-write.R
juliasilge: Yeah, let's try it. A release is still a little ways out so we can see if it is a problem in an expected way.

R/pin-read-write.R
```diff
@@ -64,6 +64,10 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
...
+#'   version if one exists (using the hash), and then **do not store** the pin
+#'   again. This argument does not check the pin metadata, only the pin
```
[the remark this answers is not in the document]
juliasilge: This would actually be better because there is an internal function already called `check_hash()`. Take a look and see if you have feedback on the new name/docs.

R/pin-read-write.R
juliasilge: Oh actually @hadley the other reason I had not set it to `TRUE` originally is it adds one additional call to `pin_meta()`. Do we think this is OK? All the API calls can be especially painful on Connect, but maybe adding yet one more set isn't much marginal difference.

tests/testthat/test-board_folder.R
```diff
@@ -86,8 +86,8 @@ test_that("generates useful messages", {
   ui_loud()
   b <- board_temp()
   expect_snapshot({
-    pin_write(b, 1:5, "x", type = "rds")
     pin_write(b, 1:5, "x", type = "rds")
     pin_write(b, 1:6, "x", type = "rds")
+    pin_write(b, 1:7, "x", type = "rds")
```
juliasilge, line 91: I did have to change a couple of tests if the default is `compare_hash = TRUE`

R/pin-read-write.R
hadley: Hmmm, I think I'd prefer an explicit `pin_exists()` check? Is there some reason not to do this?

R/pin-read-write.R
hadley: Hmmmm, I think it's probably still worth it. It definitely saves you a lot of time/space in the case where the data hasn't changed. We'll definitely need to promote that change heavily.

R/pin-read-write.R
```diff
@@ -64,6 +64,10 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #'   use the default for `board`
 #' @param tags A character vector of tags for the pin; most important for
 #'   discoverability on shared boards.
+#' @param compare_hash Compare the pin contents to the last version if one
```
hadley, line 67: I think this is an improvement, but I wondered about something even more direct like `write_if_different`? Or `write_unchanged_data`? What do you think?

juliasilge: What about `force_identical_write`?

hadley: None of these ideas really speak to me (including mine), so just go with your heart 😄

PR #740, opened 2023-05-05
Add R 🔁 Python cross compatibility tests

### LINKED ISSUE #610: Test compatibility with python pins
From discussion with @juliasilge, let's set up a github action to ensure that changes made to R pins are compatible with the current version of python. E.g. save the pins from R, and the run py pins to ensure it can read, search, list, etc..

juliasilge:
Closes #610


juliasilge:
@machow can you see if you have any feedback on how I set up this round trip 🔁 test?

PR #741, opened 2023-05-08
Make API version in metadata more consistent

### LINKED ISSUE #612: Ensure R and python pins have compatible api_version behavior
Currently, it seems like R and python pins treat api_version differently:

* R pins:
  - [Sometimes sets api_version using `0L`, `1L`](https://github.com/rstudio/pins/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/meta.R#L7-L12), but reads it as a numeric (screenshot below)
  - Has [an error message](https://github.com/rstudio/pins/blob/9e8c08123861eeef4853caa5ffd1970c9be0457e/R/meta.R#L20) that suggests version might have a format like `x.y.z` in the future
* Python pins:
  - Sets api_version as an integer (`1`), and has a type hint that suggests it expects an integer
  - Just reads api_version from the yaml, so fine reading R pins metadata, and having it be a float (e.g. `1.0`)

## Options

It seems like if we want to preserve compatibility with the current R pins, we have two safe options for api_version:

* **one part**: It follows the form `0`, `1`, `2`, etc..
* **two parts**: It allows one period, e.g. `1.1`, `2.0`

And then one crazy option--**three parts**, allowing the string `"x.y.z"`-- since in R:

* `"0" > 1` -> False
* `"1" > 1` -> False
* But we'd have to be very careful, since `"1.0" > 1` -> True

(Also I can't remember exactly how these comparisons work, so could be missing some bad behavior; but it should be okay as long as `"2.0.0" etc > 1`?)

juliasilge:
Closes #612 

This PR makes the API version (currently, possible values are 0 and 1, or it can be absent for old pins) stored in the metadata more consistent, i.e. stored and read as int.

The error message is intended to communicate about the package version (e.g. 1.1.0.9000 right now) to the user. The API stored in the metadata is more like an internal value and the package version is more like an external value. If we ever drastically changed the user-facing API again, we would bump the package version to 2.0.0, `api_version` to 2, and go from there.

### DIFF
--- a/R/meta.R
+++ b/R/meta.R
@@ -16,8 +16,8 @@ read_meta <- function(path) {
     yaml$created <- parse_8601_compact(yaml$created)
     yaml$user <- yaml$user %||% list()
   } else if (yaml$api_version > 1) {
-    abort(c(
-      paste0("Metadata requires pins ", yaml$api_version),
+    cli::cli_abort(c(
+      "Metadata requires pins {yaml$api_version}.0.0 or greater",
       i = "Do you need to upgrade the pins package?"
     ))
   }
--- a/R/meta.R
+++ b/R/meta.R
@@ -17,7 +17,7 @@ read_meta <- function(path) {
     yaml$user <- yaml$user %||% list()
   } else if (yaml$api_version > 1) {
     abort(c(
-      paste0("Metadata requires pins ", yaml$api_version, ".0.0 or greater"),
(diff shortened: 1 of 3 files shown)


juliasilge:
If we ever did need to compare the API version to the package version, we could use [`compareVersion()`](https://stat.ethz.ch/R-manual/R-devel/library/utils/html/compareVersion.html).

PR #742, opened 2023-05-08
Switch to webfakes httpbin

juliasilge:
This PR only switches over the old `pin()` tests, but when `/cache` and `/etag` are added as outlined in r-lib/webfakes#3 then I can switch over the `board_url()` tests.

### DIFF
--- a/tests/testthat/test-pin.R
+++ b/tests/testthat/test-pin.R
@@ -1,4 +1,5 @@
 skip_if_not_installed("filelock")
+skip_if_not_installed("webfakes")
 
 # main types --------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-pin.R
```diff
@@ -1,4 +1,11 @@
 skip_if_not_installed("filelock")
+skip_if_not_installed("webfakes")
+
+httpbin <- local_httpbin_app()
+httpbin_port <- httpbin$get_port()
+redact_port <- function(snapshot) {
+  snapshot <- gsub(httpbin_port, "<port>", snapshot, fixed = TRUE)
+}
```
juliasilge, line 8: In the future, I'll need to put this same bit in the `board_url()` test file. I couldn't put this in `setup.R` because webfakes is only in Suggests and it is very helpful to keep that "hard" GH action passing for pins.

PR #748, opened 2023-06-02
Add new `PINS_CACHE_PATH` env var, plus documentation

### LINKED ISSUE #611: .onLoad fails in loadNamespace on read-only filesystem (AWS lambda)
I am trying to read a pin from an Rsconnect board in a containerized AWS lambda function (runtime: R 4.1.3, pins 1.0.1).
Although I am specifying the cache directory to be on the [ephemeral storage of lambda](https://aws.amazon.com/blogs/compute/using-larger-ephemeral-storage-for-aws-lambda/) (`/tmp`) as follows: `board <- pins::board_rsconnect(server = Sys.getenv("CONNECT_SERVER"),  cache = "/tmp")` (the CONNECT_SERVER and CONNECT_API_KEY have been added as environmental variables to the lambda function configuration), still I get an error that on the loading of the namespace an attempt at creating a directory in `/home` is being made that is not writable. Is there a way to circumvent this behavior? I am loading at least 20 other packages in the runtime without any problems - it is specific to the behavior of pins:

```
Error: package or namespace load failed for ‘pins’:
.onLoad failed in loadNamespace() for 'pins', details:
call: NULL
error: [EROFS] Failed to make directory '/home/sbx_user1051': read-only file system
```

I checked  [r-pkgs on side effects on load](https://r-pkgs.org/r.html#when-you-do-need-side-effects) and in [zzz.R](https://github.com/rstudio/pins/blob/main/R/zzz.R) I noticed that the function `board_register_local` gets called `.onLoad`, without any options
 https://github.com/rstudio/pins/blob/2a1a0d89af8e9e3b696c656d5bb166502cec1ee0/R/zzz.R#L12 . Is there a way to pass `/tmp` as a cache directory here?  Could it be read from an environmental variable?

### LINKED ISSUE #747: Assignment of non-standard cache folder
Hey everybody, 
we have been using this wonderful package for a long time in our team and since we moved to Posit Workbench we are wondering how to use a shared cache on the Workbench server for pins.
In the documentation of the package and in the source code I could not find any hints that the location where the caching takes place can be customized.
However, for us it would be important to be able to use a shared cache for our pins for performance reasons (cache only needs to be built once and is then available to everyone on the server) and for disk space reasons (pins duplicates in the respective user folders just take up a large amount of disk space).

Is there any possibility or is it planned (soon) to enable this or am I missing something?

Thanks and best regards,
Janis

juliasilge:
Closes #747 

With these changes, we can now use an environment variable to specify where the cache goes:

``` r
withr::local_envvar(list("PINS_CACHE_PATH" = "~/Desktop"))
server <- pins:::rsc_server(auth = "envvar")
pins::board_cache_path(paste0("connect-", rlang::hash(server$url)))
#> ~/Desktop/connect-31a399613ac85efda77ffe6ff20c105a
```

<sup>Created on 2023-06-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This also finally fixes #611 by documenting how the env variables work, with some advice on a read-only filesystem like AWS Lambda.

### DIFF
--- a/R/board.R
+++ b/R/board.R
@@ -84,22 +84,47 @@ cache_size <- function(board) {
 
 is.board <- function(x) inherits(x, "pins_board")
 
-#' Retrieve Default Cache Path
+#' Retrieve default cache path
 #'
 #' Retrieves the default path used to cache boards and pins. Makes
-#' use of the `rappdirs` package to use cache folders
-#' defined by each OS.
+#' use of [rappdirs::user_cache_dir()] for cache folders defined by each OS.
+#' Remember that you can set the cache location for an individual board object
+#' via the `cache` argument.
 #'
 #' @param name Board name
-#' @keywords internal
+#' @details
+#' There are several environment variables available to control the location of
+#' the default pins cache:
+#'
+#' - Use `PINS_CACHE_PATH` to set the cache path for _only_ pins functions
+#' - Use `R_USER_CACHE_DIR` to set the cache path for all functions that use rappdirs
+#'
+#' On system like AWS Lambda that is read only (for example, only `/tmp` is
+#' writeable), set either of these to [base::tempdir()]. You may also need to
+#' set environment variables like `HOME` and/or `R_USER_DATA_DIR` to the
+#' session temporary directory.
+#'
 #' @examples
 #' # retrieve default cache path
 #' board_cache_path("local")
+#'
+#' # set with env vars:
+#' withr::with_envvar(
+#'     c("PINS_CACHE_PATH" = "/path/to/cache"),
+#'     board_cache_path("local")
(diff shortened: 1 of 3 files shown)


machow:
I just double checked in pins-python, and apparently it uses `PINS_CACHE_DIR` as an optional env var for overriding the cache directory. (I don't think this is documented, and probably exists for unit tests..).

https://github.com/rstudio/pins-python/blob/7ded06943c6ea3236e5da4443a3869073d66aeb1/pins/config.py#L8

juliasilge:
Thanks to both!

@machow I changed the env var here so it matches what pins for Python has. 👍

PR #749, opened 2023-06-07
Add Google Drive board

### LINKED ISSUE #728: Consider board based on googledrive R package
Let's consider a new pins board based on the [googledrive](https://googledrive.tidyverse.org/) package. 

🎯 What should its name be? `board_googledrive()`? That's a little long, maybe?

juliasilge:
Closes #728 

Wow, using Google Drive for this is _not_ speedy 😳


juliasilge:
We should definitely not run the tests for `board_gdrive()` in CI; they take basically INFINITY amount of time.

hadley (approved):
Any hope of speeding it up by caching drive ids somewhere?

juliasilge:
> Any hope of speeding it up by caching drive ids somewhere?

I don't believe so. There's no "extra" mapping between drive ID and something else that we can cache somewhere (like the Connect mappings from GUID to content name).

Just for clarity here, the normal pins caching mechanism does work as expected here, so you don't re-download pin context if it hasn't changed. However, even the process of looking up the pin metadata to _check_ this feels sluggish, however.

Issue #481, opened 2021-08-17
How to create a board


hadley:
Write it up, at least for my own use.

jrosell:
Maybe sftp/scp too.

juliasilge:
What if we documented this as a `README` inside of `R/`, instead of a user-facing vignette? [parsnip has one of these](https://github.com/tidymodels/parsnip/blob/main/R/README.md), and it seems like a good, low-lift option for this kind of developer-facing info.

hadley:
I don't love it, just because I personally would have no idea to look there.

juliasilge:
Thanks to @edgararuiz for pointing out that you also need to create:

- `pin_delete()`
- `pin_version_delete()` for boards with prefixes

PR #751, opened 2023-06-14
New "create a new board" vignette

### LINKED ISSUE #737: Enabling extension of pins to other platforms
Yes, definitely, I just want to make sure the extension interface is solid before encouraging others to use it.

_Originally posted by @hadley in https://github.com/rstudio/pins-r/issues/349#issuecomment-847360561_
 
Would it be possible to resurrect pins-user's ability to develop extensions for data repositories beyond some of the ones that are covered (e.g. S3, GCS, Azure etc.)?

*Context*: pre 1.0 version, pins used to provide support for users to write extensions for pinning to not-broadly-supported boards. In conversation with @hadley ~ 2 years ago, my understanding was that, the ability to extend pins to other data repos (boards) will be restored. For big organizations with variety of less common data platforms, the uniformity that pins could bring (no matter which back-end data platform) is of great value.

juliasilge:
Closes #737 as well as #481

I got a start on this draft while working on #749 and used the Google Drive board as the example. It's a pretty good example, mostly because the googledrive R package is so nice.

One thing this vignette highlights is that a lot of functions needed to create these methods are internal, like `version_setup()`, `check_pin_exists()`, and so forth, are all unexported. Should we export them? Or show them being used like `pins:::check_pin_exists()` in this vignette?

### DIFF
--- a/vignettes/create-new-board.Rmd
+++ b/vignettes/create-new-board.Rmd
@@ -1,0 +1,205 @@
+---
+title: "How to create a new board"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{create-new-board}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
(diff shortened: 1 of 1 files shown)


vignettes/create-new-board.Rmd
```diff
@@ -0,0 +1,205 @@
...
+
+  name
+}
+```
+
+This vignette presents writing these methods as if the process is perfectly linear, but you'll probably revisit earlier steps as you continue creating your methods. To make sure your methods all work as expected, considering using these pins testing functions with a test board `board`: 
+
+- `test_api_basic(board)`
+- `test_api_versioning(board)`
+- `test_api_meta(board)`
```
juliasilge, line 197: These are unexported too.

PR #752, opened 2023-06-21
Try out tmate session

juliasilge:
Debugging new failures for "check board" GH action

### DIFF
--- a/.github/workflows/check-boards.yaml
+++ b/.github/workflows/check-boards.yaml
@@ -41,6 +41,10 @@ jobs:
           extra-packages: any::rcmdcheck
           needs: check
 
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 30
+
       - uses: r-lib/actions/check-r-package@v2
         with:
           error-on: '"note"'


juliasilge:
So when ssh-ing into the action to check things out,

- Tests seems fine if I run them from within R
- Running tests like `R -e "devtools::test(filter = 'board_azure_blob')"` can fail

I am noticing that the earlier failures for this action involved between 2 and 14 failures, so there is some flakiness involved.

I believe this is what is failing, within `test_api_meta()` and specifically "can round-trip pin metadata":

```r
name <- local_pin(board, 1, title = "xxx-a1", description = "xxx-a2")
    
# change content so hash changes
pin_write(board, 2, name, title = "xxx-b1", description = "xxx-b2")
```

That second write looks like it is not happening, and the metadata is still un-updated.

juliasilge:
Looks like it's the Azure boards that are intermittently failing on that second write, as #753 solves the problem

juliasilge:
WELP, the problem is still the Azure tests. I have _no idea_ why #753 isn't actually skipping the Azure tests right now, to the point that I feel like I'm losing my mind.

PR #753, opened 2023-06-21
Skip Azure tests on CI

juliasilge:
It _might_ be just the Azure tests that are a problem?

### DIFF
--- a/.github/workflows/check-boards.yaml
+++ b/.github/workflows/check-boards.yaml
@@ -18,7 +18,7 @@ jobs:
       # needed for modern board tests
       ## PINS_AWS_ACCESS_KEY: ${{ secrets.PINS_AWS_ACCESS_KEY}}
       ## PINS_AWS_SECRET_ACCESS_KEY: ${{ secrets.PINS_AWS_SECRET_ACCESS_KEY}}
-      ## PINS_GCS_PASSWORD: ${{ secrets.PINS_GCS_PASSWORD }}
+      PINS_GCS_PASSWORD: ${{ secrets.PINS_GCS_PASSWORD }}
 
       # needed for legacy board tests
       TEST_AWS_BUCKET: "pins-github-tests"
--- a/.github/workflows/check-boards.yaml
+++ b/.github/workflows/check-boards.yaml
@@ -16,9 +16,6 @@ jobs:
       R_KEEP_PKG_SOURCE: yes
 
       # needed for modern board tests
-      PINS_AZURE_SAS: ${{ secrets.PINS_AZURE_SAS }}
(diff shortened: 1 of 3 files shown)


juliasilge:
Yeah, I think the Azure tests are flaky for some reason. They run fine for me locally. FWIW it is only `test_api_meta()` and specifically "can round-trip pin metadata" that are the problem, not the whole suite of tests.

juliasilge:
@hadley I do want to merge this in, I think, because this issue has not resolved. I don't know what is going on with the Azure tests, as they run fine locally.

juliasilge:
@isabelizimm have you experienced any problems with the Azure tests on pins for Python in CI? I believe we are using the same credentials. You can see the problems [here](https://github.com/rstudio/pins-r/actions/runs/5813389631/job/15761317623?pr=764) and [here](https://github.com/rstudio/pins-r/actions/runs/5351231834/job/14487844900); it's the tests for updating metadata like the description.

hadley (approved):
Makes sense to me. I assume these are already skipped on CRAN through a different mechanism.

juliasilge:
Yes, the cloud board tests are skipped on CRAN via the `skip_if_missing_envvars()` helper:

https://github.com/rstudio/pins-r/blob/ff30f5955830768b9bc7843aba50d13167141b21/R/board_azure.R#L87

I guess a different option than this PR would be to take the Azure env vars out of the repo and only have them locally. This is what we do for MS365 and Google Drive tests (just never add credentials to the repo to start with).

isabelizimm:
I just ran the [pins-python tests,](https://github.com/rstudio/pins-python/actions/runs/5828079536) and they didn't have any issues on Azure 🤔

juliasilge:
OK, if I turn off _all_ the cloud board tests, the [check-boards action](https://github.com/rstudio/pins-r/actions/runs/5836511972/job/15830241479) passes. ✅

juliasilge:
GCS tests FAIL 🙅🏻‍♀️

I wonder if it is something about checking the hash on these cloud boards.

juliasilge:
In the end, it was a problem with the test related to the timestamp, for versions being written very close together. Took me forever to find that out, because of not being able to replicate it except in very specific circumstances! The changes to the test in #765 make it more robust, and we don't need to turn anything off.

machow:
In case it's useful down the road, for testing on windows I ended up having to sleep like 300ms between pin writes (and file touches via pin_read or pin_meta), in order to get all the tests to pass consistently.

https://github.com/rstudio/pins-python/blob/ee96f2d85fc5a4633678c1b9cb49fbac1a73221c/pins/tests/test_cache.py#L15-L27

PR #754, opened 2023-06-22
Fix uses of `possibly()`

juliasilge:
Somehow I was totally unaware of [this](https://purrr.tidyverse.org/reference/faq-adverbs-export.html), despite it being linked from the `possibly()` docs.

### DIFF
--- a/R/board_gdrive.R
+++ b/R/board_gdrive.R
@@ -172,7 +172,7 @@ required_pkgs.pins_board_gdrive <- function(x, ...) {
 # Helpers -----------------------------------------------------------------
 
 possibly_drive_ls <- function(...) {
-  purrr::possibly(googledrive::drive_ls)(...)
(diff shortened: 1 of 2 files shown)


juliasilge:
The "check-boards" action fails because of the problems outlined in #753.

R/board_gdrive.R
```diff
@@ -171,10 +171,13 @@ required_pkgs.pins_board_gdrive <- function(x, ...) {
 
 # Helpers -----------------------------------------------------------------
 
+possibly_drive_ls <- function(...) {
+  purrr::possibly(googledrive::drive_ls)(...)
+}
```
hadley, line 176: If you're going to write a wrapper, I think you might as well just inline `possibly()`, i.e.

```suggestion
possibly_drive_ls <- function(...) {
  tryCatch(googledrive::drive_ls(...), error = function(err) NULL)
}
```

PR #765, opened 2023-08-11
Use `force_identical_write` to update metadata in test

juliasilge:
The problems with testing on the cloud boards have been making my feel like I am LOSING MY MIND, but maybe it is related to this.


juliasilge:
Oh my gosh, it was the TIMESTAMP ALL ALONG ☠️ 

Closes #753

Issue #777, opened 2023-09-01
Error when using pin_upload() in a pin board created by board_gdrive()


gorkang:
Hi there. I am trying to use a Shared Google Drive folder to create pin boards to upload and download files. 

If I do that with a local folder with `board_folder()`, everything works fine. But I can't make `board_gdrive()` work with a shared Google Drive folder.

The containing folders for the pin board are created, but the file never uploads, and I get some errors and warnings.

``` r
drive_link = "https://drive.google.com/drive/folders/1GNs7jJeKqVgv7qA5nRcNzGHbCejOZpDh"
googledrive::drive_auth(email = "myemail@gmail.com")

gdrive_board = pins::board_gdrive(drive_link)
gdrive_board |> pins::pin_list()
#> character(0)

gdrive_board |> pins::pin_upload(paths = "~/Downloads/999.zip", name = "First_pin")
#> Creating new version '20230901T115649Z-db9b5'
#> Warning: Unknown or uninitialised column: `path`.
#> Error in if (is.null(name) && !has_slash(path) && drive_path_exists(append_slash(path))) {: missing value where TRUE/FALSE needed
gdrive_board |> pins::pin_list()
#> [1] "First_pin"

gdrive_board |> pins::pin_upload(paths = "~/Downloads/999.zip", name = "First_pin")
#> Warning: Unknown or uninitialised column: `path`.
#> Error in `as_parent()`:
#> ! Parent specified via `path` is invalid:
#> ✖ Does not exist.
#> Backtrace:
#>      ▆
#>   1. └─pins::pin_upload(...)
#>   2.   ├─pins::pin_store(board, name, paths, meta, ...)
#>   3.   └─pins:::pin_store.pins_board_gdrive(...)
#>   4.     └─pins:::version_setup(board, name, version_name(metadata), versioned = versioned)
#>   5.       └─pins::pin_versions(board, name)
#>   6.         ├─pins:::pin_versions_modern(board, name, ...)
#>   7.         └─pins:::pin_versions.pins_board_gdrive(board, name, ...)
#>   8.           ├─pins:::version_from_path(sort(googledrive::drive_ls(path)$name))
#>   9.           │ └─pins:::versions_template(x)
#>  10.           ├─base::sort(googledrive::drive_ls(path)$name)
#>  11.           └─googledrive::drive_ls(path)
#>  12.             └─googledrive:::as_parent(path)
[...]

PR #780, opened 2023-09-03
Use a dribble to define the upload destination path

gorkang:
This is a potential fix to #777 

Instead of passing a full path to `drive_upload`, with this pr we first create a dribble with the destination folder, and then use it as a path. 

In my computer, it works with a single or multiple files, uploading to the main google drive folder or to  folder with a long path.

For example, now this works without any issues:
```
drive_link = "https://drive.google.com/drive/folders/1_jiG0VtENy5QcLRziN7DM_zQD4egwGuN"
googledrive::drive_auth(email = "myemail@gmail.com")
gdrive_board = pins::board_gdrive(drive_link)
gdrive_board |> pins::pin_upload(paths = "~/Downloads/999.zip", name = "Id_pin")
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 
 * Added `board_deparse` for `board_url()` (#774).
 
+* Fixed how `board_gdrive()` makes version directories (#780, @gorkang).
+
 
 # pins 1.2.1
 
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
Thank you so much for this fix @gorkang! 🙌 I tested it out locally and this works great for me with my own Google Drive. 

(Side note: We don't run the Google Drive or MS OneDrive tests in CI because they are _so_ slow and authentication is such a hassle.)

The round trip R 🔄 Python tests are failing right now because of a Python package dependency issue, but we'll get that sorted and then I'll merge this in.

juliasilge:
We got the Python problem fixed in https://github.com/rstudio/pins-python/pull/212. 👍

PR #787, opened 2023-09-20
Add "write" to warning about using full name when referencing pin

### LINKED ISSUE #786: Fix confusing message about "when reading a pin" for `pin_write()`
If you don't include the username when writing a pin to Connect, it gives a confusing message:

``` r
library(pins)
board <- board_connect()
#> Connecting to Posit Connect 2023.07.0 at <https://colorado.posit.co/rsc>
board |> pin_write(10:20, "nice-numbers", type = "json")
#> ! Please use full name when reading a pin: "julia.silge/nice-numbers", not "nice-numbers".
#> Writing to pin 'julia.silge/nice-numbers'
```

<sup>Created on 2023-09-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Notice:

> when reading a pin

but we were writing! It happens in `rsc_content_find()`, which is used a lot and not only when reading. Probably should update this message:

https://github.com/rstudio/pins-r/blob/1a1e450ea7bff945fce7a94dc9fb87d2f778f703/R/board_connect.R#L474

rkb965:
Closes #786

See discussion https://github.com/posit-conf-2023/vetiver/discussions/31

### DIFF
--- a/R/board_connect.R
+++ b/R/board_connect.R
@@ -471,7 +471,7 @@ rsc_content_find_live <- function(board, name, version = NULL, warn = TRUE) {
     name$full <- paste0(owner, "/", name$name)
 
     if (warn) {
-      cli::cli_alert_warning("Please use full name when reading a pin: {.val {name$full}}, not {.val {name$name}}.")
+      cli::cli_alert_warning("Please use full name when reading or writing a pin: {.val {name$full}}, not {.val {name$name}}.")
     }
     selected <- json[[1]]
   } else {
--- a/R/board_connect.R
+++ b/R/board_connect.R
@@ -471,7 +471,7 @@ rsc_content_find_live <- function(board, name, version = NULL, warn = TRUE) {
     name$full <- paste0(owner, "/", name$name)
 
     if (warn) {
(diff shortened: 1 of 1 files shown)


juliasilge:
This function is used not only when reading or writing pin contents, but also when calling `pin_meta()` and `pin_exists()`. Instead of just _adding_ "write", can we change the message to be more like the one here:

https://github.com/rstudio/pins-r/blob/1a1e450ea7bff945fce7a94dc9fb87d2f778f703/R/board_connect.R#L467

R/board_connect.R
```diff
@@ -471,7 +471,7 @@ rsc_content_find_live <- function(board, name, version = NULL, warn = TRUE) {
     name$full <- paste0(owner, "/", name$name)
 
     if (warn) {
-      cli::cli_alert_warning("Please use full name when reading a pin: {.val {name$full}}, not {.val {name$name}}.")
+      cli::cli_alert_warning("Please use full name when reading or writing a pin: {.val {name$full}}, not {.val {name$name}}.")
```
juliasilge, line 474:
```suggestion
      cli::cli_alert_warning("Use a fully specified name including user name: {.val {name$full}}, not {.val {name$name}}.")
```

juliasilge (approved):
Thank you so much @rkb965! 🚀

PR #792, opened 2023-09-29
Change the `pin_write()` function signature, moving the dots

juliasilge:
We've been talking about `...` being in the wrong place for quite a while, and we think it's time to rip the band-aid off for a better experience for people. 

Technically we "should" put the dots _before_ `name` but it is very very very common for folks to provide `name` by position (without `name = `). We do it in all our docs and tests, and I believe it is very common with users as well. I would like to propose that we move the dots to between `name` and `type`, as it will not be too disruptive for users and we'll be able to get to our better situation of moving around or adding args after the dots.

The new error that folks see if they are passing by position is this:

``` r
library(pins)
board <- board_temp()
board |> pin_write(10:20, "nice-numbers", "json", "my great title")
#> Error in `pin_write()`:
#> ! Arguments after the dots `...` must be named, like `type = "json"`.
#> Backtrace:
#>     ▆
#>  1. └─pins::pin_write(board, 10:20, "nice-numbers", "json", "my great title")
#>  2.   └─cli::cli_abort("Arguments after the dots `...` must be named, like {.code type = \"{dots[[1]]}\"}.") at pins-r/R/pin-read-write.R:85:4
#>  3.     └─rlang::abort(...)
```
<sup>Created on 2023-09-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The message doesn't make quite as much sense if people are using a mix of calling by name and position, but surely this must be rare:

``` r
library(pins)
board <- board_temp()
board |> pin_write(10:20, name = "nice-numbers", type = "json", "my great title")
#> Error in `pin_write()`:
#> ! Arguments after the dots `...` must be named, like `type = "my great
#>   title"`.
#> Backtrace:
#>     ▆
#>  1. └─pins::pin_write(board, 10:20, name = "nice-numbers", type = "json", "my great title")
#>  2.   └─cli::cli_abort("Arguments after the dots `...` must be named, like {.code type = \"{dots[[1]]}\"}.") at pins-r/R/pin-read-write.R:85:4
#>  3.     └─rlang::abort(...)
```

<sup>Created on 2023-09-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

At least they are still getting an error with some helpful info, even if it isn't as fully helpful.

### DIFF
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -71,15 +71,19 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
 #' @export
 pin_write <- function(board, x,
                       name = NULL,
+                      ...,
                       type = NULL,
                       title = NULL,
                       description = NULL,
                       metadata = NULL,
                       versioned = NULL,
                       tags = NULL,
-                      ...,
                       force_identical_write = FALSE) {
   check_board(board, "pin_write", "pin")
+  dots <- list2(...)
+  if (!is_empty(dots) && is.null(names(dots[1]))) {
+    cli::cli_abort('The {.arg type} argument must be named, like {.code type = "{dots[[1]]}"}.')
+  }
 
   if (is.null(name)) {
     name <- enexpr(x)
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -82,7 +82,7 @@ pin_write <- function(board, x,
   check_board(board, "pin_write", "pin")
   dots <- list2(...)
   if (!is_empty(dots) && is.null(names(dots[1]))) {
-    cli::cli_abort('The {.arg type} argument must be named, like {.code type = "{dots[[1]]}"}.')
+    cli::cli_abort('Arguments after the dots `...` must be named, like {.code type = "{dots[[1]]}"}.')
   }
 
   if (is.null(name)) {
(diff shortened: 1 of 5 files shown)


R/pin-read-write.R
```diff
@@ -71,15 +71,19 @@ pin_read <- function(board, name, version = NULL, hash = NULL, ...) {
...
                       title = NULL,
                       description = NULL,
                       metadata = NULL,
                       versioned = NULL,
                       tags = NULL,
-                      ...,
                       force_identical_write = FALSE) {
   check_board(board, "pin_write", "pin")
+  dots <- list2(...)
+  if (!is_empty(dots) && is.null(names(dots[1]))) {
```
hadley, line 84: I don't think the logic is quite right here as `names` will be `NULL` if all the elements are unnamed, but if there's one named element, it'll be `""`.

I don't know if `...names()` is available in old enough R for us to use it, but you could also try something like this:

```R
if (!missing(...) && !(is.null(...names())) || ...names()[[1]] == ""))
```

hadley (approved):
Your reasoning makes sense to me.

R/pin-read-write.R
juliasilge: Looks like we do not yet have `...names()` in oldrel-3 and oldrel-4.

PR #795, opened 2023-10-18
Add new `urls` to metadata

juliasilge:
One thing we thought about when listening to JD Long's keynote talk at posit::conf last month was the value of making sure it is as easy as possible connect datasets to documentation. This PR is inspired by those thoughts and adds a new option for storing `urls` (a character vector) along with your dataset.

It's used like this:

``` r
library(pins)

b <- board_connect()
#> Connecting to Posit Connect 2023.07.0 at <https://colorado.posit.co/rsc>
b |> pin_write(
    c(200, 404, 503),
    "http-code-numbers",
    urls = c("https://httbey.com/", "https://http.cat/"),
    force_identical_write = TRUE
)
#> Guessing `type = 'rds'`
#> Writing to pin 'julia.silge/http-code-numbers'
```

<sup>Created on 2023-10-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I added the URLs to the Connect preview, which you can see here:
https://colorado.posit.co/rsc/httpbey-pin/

Any thoughts on a better way to present that?

### DIFF
--- a/R/pin-meta.R
+++ b/R/pin-meta.R
@@ -12,6 +12,7 @@
 #'   * `$title` - pin title
 #'   * `$description` - pin description
 #'   * `$tags` - pin tags
+#'   * `$urls` - URLs for more info on pin
 #'   * `$created` - date this (version of the pin) was created
 #'   * `$api_version` - API version used by pin
 #'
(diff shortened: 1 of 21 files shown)


R/pin-meta.R
```diff
@@ -97,12 +98,13 @@ empty_local_meta <- local_meta(x = NULL, name = NULL, dir = NULL)
 
 test_api_meta <- function(board) {
   testthat::test_that("can round-trip pin metadata", {
-    name <- local_pin(board, 1, title = "title", description = "desc", metadata = list(a = "a"), tags = c("tag1", "tag2"))
+    name <- local_pin(board, 1, title = "title", description = "desc", metadata = list(a = "a"), tags = c("tag1", "tag2"), urls = c("https://posit.co/"))
```
hadley, line 101:
```suggestion
    name <- local_pin(board, 1, title = "title", description = "desc", metadata = list(a = "a"), tags = c("tag1", "tag2"), urls = "https://posit.co/")
```

inst/preview/index.html
```diff
@@ -46,6 +46,11 @@ <h3>{{{pin_name}}}</h3>
        {{/pin_metadata}}
      </section>
 
+    <section style="{{url_preview_style}}">
+      <h3>Learn more about this pin at:</h3>
+      <p>{{{urls}}}</p>
+    </section>
```
hadley, line 52:
```suggestion
    {#urls}
    <section>
      <h3>Learn more about this pin at:</h3>
      <p>{{{.}}}</p>
    </section>
    {/urls}
```
?

R/pin-meta.R
```diff
@@ -133,6 +135,9 @@ test_api_meta <- function(board) {
     testthat::expect_snapshot_error(
```
hadley, line 135: I feel like this test could be run just once for `pin_write()` since the code doesn't vary across backends?

hadley (approved):
I like it!

R/pin-meta.R
juliasilge: Ah, great point, yep 👍

Issue #796, opened 2023-10-29
Connect preview html: include Python code too?


nealrichardson:
If you make a pin with the R package, you get https://github.com/rstudio/pins-r/blob/main/inst/preview/index.html created with it, which tells the Connect user how to load the data. But it only has R code there, even though you can totally use pins-python to read it (right?) 

It would be nice if the code example was tabbed and had Python and R example code. (Assuming you agree, there should also be the same done to pins-python, https://github.com/rstudio/pins-python/blob/main/pins/rsconnect/html/index.html is Python only).

juliasilge:
Yeah, thanks for bringing this back up; you absolutely can roundtrip a pin through R and Python, as long as you use a file format that is amenable (i.e. not RDS, not joblib, etc).

We did talk about this when we first started the Python package. The main challenge, as I remember it, is running the Python `board_deparse()` from R and vice versa. Right now, it is pretty simple. Here is what Python would need to do for R:

https://github.com/rstudio/pins-r/blob/f6c6f50969e0ab9be091c838075cd619570f70b3/R/board_connect.R#L365-L367

but we are tracking changes that need to be made in #726. Also doesn't look so bad for [what R would need to do for Python](https://github.com/rstudio/pins-python/blob/5d81141ea13bc766fe7b6b42d291b0e9d7264d28/pins/constructors.py#L56).

We had other concerns at the time about trying to stay consistent with different versions of the R and Python packages but I am not so worried about that now, as that basic read/write interface is pretty solid. Also, we would need to check the metadata before generating the HTML to know if this file _can_ be read from the other language.

juliasilge:
We've got this implemented in the R package now and I have opened up rstudio/pins-python#222 to get it done in Python.

PR #806, opened 2023-11-10
Add Python code to Connect preview HTML

juliasilge:
Addresses #796

We discussed a button to switch between the R / Python code, but we need to _not_ show the code at all for `.rds` pins so I think it's better to just do it with another `<section>`. This is my current proposal:

Screenshot 2023-11-09 at 6 07 29 PM

You can see a [published version here](https://colorado.posit.co/rsc/content/a444e2ba-1d26-4945-99ce-7d184a82385d).

### DIFF
--- a/inst/preview/index.html
+++ b/inst/preview/index.html
@@ -62,10 +62,13 @@ pin_read(board, "{{pin_name}}")</code></pre>
         hljs.initHighlightingOnLoad();
       </script>
     </section>
-    
+
     <section style="{{show_python_style}}">
       <h3>Python Code</h3>
         <pre id="pin-python" class="pin-code"><code class="python">import pins
+from dotenv import load_dotenv
+load_dotenv()
+
 board = pins.board_connect()
 board.pin_read(board, "{{pin_name}}")</code></pre>
     </section>
--- a/inst/preview/index.html
+++ b/inst/preview/index.html
@@ -53,7 +53,7 @@
 
     <section>
       <h3>R Code</h3>
-        <pre id="pin-r" class="pin-code-r"><code class="r">library(pins)
+        <pre id="pin-r" class="pin-code"><code class="r">library(pins)
 board <- {{{board_deparse}}}
 pin_read(board, "{{pin_name}}")</code></pre>
       <script type="text/javascript">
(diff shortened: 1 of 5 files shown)


inst/preview/index.html
```diff
@@ -52,14 +52,22 @@ <h3>Learn more about this pin at:</h3>
...
+        hljs.registerLanguage("python", highlight_python);
+        hljs.initHighlightingOnLoad();
+      </script>
+    </section>
+    
+    <section style="{{show_python_style}}">
+      <h3>Python Code</h3>
+        <pre id="pin-python" class="pin-code-python"><code class="python">import pins
+board = pins.board_connect()
+board.pin_read(board, "{{pin_name}}")</code></pre>
```
juliasilge: Is this the best option for Python code? I believe this is what people should use if they are publishing code to Connect, right? Because the env vars are provided?

inst/preview/index.html
```diff
@@ -33,7 +33,7 @@
        <h3>{{{pin_name}}}</h3>
        {{#pin_metadata}}
        <p>
-         {{#date}}<b>Last updated:</b> {{{.}}} &bull;{{/date}}
+         {{#date}}<b>Last updated from R:</b> {{{.}}} &bull;{{/date}}
```
juliasilge, line 36: The Python package can say:

> Last updated from Python:

inst/preview/index.html
```diff
@@ -52,14 +52,22 @@ <h3>Learn more about this pin at:</h3>
...
+        hljs.registerLanguage("python", highlight_python);
+        hljs.initHighlightingOnLoad();
+      </script>
+    </section>
+    
+    <section style="{{show_python_style}}">
+      <h3>Python Code</h3>
+        <pre id="pin-python" class="pin-code"><code class="python">import pins
+board = pins.board_connect()
+board.pin_read(board, "{{pin_name}}")</code></pre>
```
juliasilge, line 70: Is this the best option for Python code? I believe this is what people should use if they are publishing code to Connect, right? Because the env vars are provided?

isabelizimm: you'll need to bring the environment variables into your session somehow, but once they are there, you won't have to add any special arguments. something like this would work ⏬ 

```
import pins
from dotenv import load_dotenv
load_dotenv()

board = pins.board_connect()
```

inst/preview/index.html
isabelizimm: I like this format for a subtle way to be able to tell what language the data is coming from!

juliasilge:
You can see the new version of Python code [now here](https://colorado.posit.co/rsc/content/a444e2ba-1d26-4945-99ce-7d184a82385d/).

PR #808, opened 2023-11-15
Use correct classed error for "same name, different user" case

### LINKED ISSUE #807: Problem looking up pins on Connect with duplicate names
On a Connect server, we have problems looking up a pin when different users have pins with the same name. This is true even when you use the fully qualified name:

``` r
library(pins)
b <- board_connect()
#> Connecting to Posit Connect 2023.10.0 at <https://colorado.posit.co/rsc>
b |> pin_write(sample(1:100, 10), "julia.silge/test", type = "json")
#> Error in `rsc_content_find()`:
#> ! Can't find pin named 'test' with owner 'julia.silge'
#> Backtrace:
#>     ▆
#>  1. └─pins::pin_write(b, sample(1:100, 10), "julia.silge/test", type = "json")
#>  2.   ├─pins::pin_store(...)
#>  3.   └─pins:::pin_store.pins_board_connect(...)
#>  4.     ├─base::tryCatch(...)
#>  5.     │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>  6.     │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  7.     │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
#>  8.     └─pins:::rsc_content_find(board, name, warn = FALSE)
#>  9.       └─rlang::abort(...)
```

<sup>Created on 2023-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

juliasilge:
Closes #807 

When a user was trying to write a new pin that had the same name as an existing pin belonging to a different user, we weren't correctly adding `class = "pins_pin_missing"` to get the right behavior in the `tryCatch()`:

https://github.com/rstudio/pins-r/blob/ed3ac3b58943474c3523d6697379e93ca01e0f87/R/board_connect.R#L276-L278

This PR adds the correct classed condition.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pins (development version)
 
+* Fixed a bug in how pins with the same name but different owners on Posit Connect were identified (#808).
+
 # pins 1.3.0
 
 ## Breaking changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,12 @@
-# pins (development version)
+# pins (development version, to be released as 1.4.0)
+
+## Lifecycle changes
+
+* Changed the function signature of `pin_upload()` to be consistent with `pin_write()` i.e. arguments like `tags` must be passed by name and not position (#809).
+
+## Other improvements
+
+* Added example Python code to pin previews for Posit Connect (#806).
 
 * Fixed a bug in how pins with the same name but different owners on Posit Connect were identified (#808).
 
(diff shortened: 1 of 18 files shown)


juliasilge:
@stevenolen Would you be up for installing this PR and trying again the code that you had trouble with? You can install with `devtools::install_github("rstudio/pins-r@fix-error-class")`.

This PR should fix the situation where you, for example, write your own pin named "steve.nolen/test". You _will_ still see an error if you try to write a pin named "test"; I think we want to keep that an error for now.

PR #809, opened 2023-11-15
Update `pin_upload()` args

### LINKED ISSUE #805: Update `pin_upload()` args to be consistent with `pin_write()`
There have been some improvements to the arguments of `pin_write()` lately, but `pin_upload()` has fallen behind in those improvements, as pointed out in #804.

It's also worth taking a look at the `pin_download()` / `pin_read()` pair.

juliasilge:
Closes #805 

After this PR, the args for `pin_write()` are now consistent with `pin_upload()`. There were no changes needed for `pin_read()` / `pin_download()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pins (development version)
 
-## Breaking changes
+## Lifecycle changes
 
 * Changed the function signature of `pin_upload()` to be consistent with `pin_write()` i.e. arguments like `tags` must be passed by name and not position (#809).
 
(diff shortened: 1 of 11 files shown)


hadley (approved):

NEWS.md
```diff
@@ -1,4 +1,10 @@
-# pins (development version)
+# pins (development version, to be released as 1.4.0)
+
+## Breaking changes
```
hadley, line 3: I've started calling these sections "Lifecycle changes" which is a bit less harsh than breaking, but still gets the point across.

PR #811, opened 2023-11-17
Improve error message for problems with URLs

### LINKED ISSUE #810: Improved error reporting for redirected mistyped pins in board_connect_url
Disclaimer: I'm not sure whether I should log this with Connect rather than with pins-r 

If I have a pin with a vanity url, but request that vanity url without a trailing slash, then I get a confusing error message about "Pin does not declare file type" - from https://github.com/rstudio/pins-r/blob/a3e52c78e799d24359b72dffd4cac33a7bc0fe98/R/pin-read-write.R#L230

It might be helpful if this reported a message like "incorrect url - have you tried adding a trailing slash" instead - or perhaps it could even return the data?

Reproduced using:
```
library(pins)
board <- board_connect_url(
  list(
    mypin_works = "https://my.connect.com/test/data_cs/",
    mypin_not_works = "https://my.connect.com/test/data_cs"
  ),
  headers = connect_auth_headers(key = Sys.getenv("MY_KEY")))
)
data1 <- pin_read(board, "mypin_works")
data2 <- pin_read(board, "mypin_not_works")
```

If I look at the http transfer... I see that the "mypin_works" requests `GET /test/data_cs/data.txt` while "mypin_not_works" requests `GET /test/data_cs` which gets redirected to `/test/data_cs/` which returns the Connect HTML page

```
httr::with_verbose({
  pin_read(board, "mypin_not_works")
}, data_in=TRUE, ssl=TRUE)
```

image

juliasilge:
Closes #810 

We now see a message like this when we end up in a bad URL situation, such as trying `pin_read()` on a full path or forgetting a trailing slash for a vanity URL:

``` r
library(pins)
board <- board_connect_url(c(
  numbers = "https://colorado.posit.co/rsc/some-nice-numbers"
))
board |> pin_read("numbers")
#> Error in `object_read()`:
#> ! Cannot automatically read pin:
#> • Is your pin specified as a full path? Retrieve it with `pin_download()`
#> • Is your pin specified via a URL that is not a full path, such as a Posit
#>   Connect vanity URL? Remember to include a trailing slash `/`
#> Backtrace:
#>     ▆
#>  1. └─pins::pin_read(board, "numbers")
#>  2.   └─pins:::object_read(meta)
#>  3.     └─cli::cli_abort(...)
#>  4.       └─rlang::abort(...)
```

<sup>Created on 2023-11-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

@slodge do you have any feedback on this?

### DIFF
--- a/R/board_connect_url.R
+++ b/R/board_connect_url.R
@@ -7,9 +7,9 @@
 #' `board_connect_url()` is read only, and does not support versioning.
 #'
 #' @param vanity_urls A named character vector of
-#'   [Connect vanity URLs](https://docs.posit.co/connect/user/content-settings/#custom-url).
-#'   This board is read only, and the best way to write to a pin on Connect is
-#'   [board_connect()].
+#'   [Connect vanity URLs](https://docs.posit.co/connect/user/content-settings/#custom-url),
+#'   including trailing slash. This board is read only, and the best way to write to a pin 
+#'   on Connect is [board_connect()].
 #' @family boards
 #' @inheritParams new_board
 #' @inheritParams board_url
--- a/R/pin-read-write.R
+++ b/R/pin-read-write.R
@@ -226,9 +226,10 @@ object_read <- function(meta) {
       joblib = abort("'joblib' pins not supported in R"),
       csv = utils::read.csv(path),
       qs = read_qs(path),
-      file = abort(c(
-        "Pin does not declare file type so can't be automatically read",
-        i = "Retrieve uploaded paths with `pin_download()`"
+      file = cli_abort(c(
+        "Cannot automatically read pin:",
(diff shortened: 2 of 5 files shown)


slodge:
Error message reads much better ... I can't promise I'll read it :) 

... but I read the previous one - so there's hope!

I'd guess there was "guessing" the code could have done in the connect_url case - e.g. looking for a file extension, detecting the html response, etc... but I agree that'd be overkill for the case where the user (me!) had typed the url wrong :+1:

Thanks!

juliasilge:
Most of the basic URL parsing tools out there (`urltools::url_parse()` and `httr::parse_url()`) don't really give us directly what we'd need to distinguish between use cases, so I do think it's better to improve the error message rather than try for more guessing. 

I appreciate the report a lot -- thanks again! 🙌

PR #814, opened 2023-12-11
Clarify authentication for Connect

### LINKED ISSUE #813: Clarify instructions of board_connect()
The description of `board_connect()` reads

> To use a Posit Connect board, you need to first authenticate. The easiest way to do so is by launching **Tools** - **Global Options** - **Publishing** - **Connect**, and follow the instructions.

I was confused for a moment before realizing that those steps should be taken from the RStudio IDE. I think it would be nice to mention that

juliasilge:
Closes #813

### DIFF
--- a/R/board_connect.R
+++ b/R/board_connect.R
@@ -2,8 +2,8 @@
 #'
 #' @description
 #' To use a Posit Connect board, you need to first authenticate. The easiest
-#' way to do so is by launching **Tools** - **Global Options** -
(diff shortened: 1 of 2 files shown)


juliasilge:
@EmilHvitfeldt do you have any further suggestions on this advice?

(FWIW I actually think people should mostly use environment variables to authenticate to Connect but using the IDE could arguably be said to be easier.)

EmilHvitfeldt:
Looks good. Agree on the use of environment variables!

PR #817, opened 2024-01-05
Update docs, especially about concurrent writes

juliasilge:
Based on feedback during a recent pins stakeholder meeting, let's further clarify what and how to write objects/data as pins.

If we think this language is good to go, I will do a PR to pins for Python as well as the Connect docs.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -29,10 +29,6 @@ The pins package publishes data, models, and other R objects, making it easy to 
 You can pin objects to a variety of pin *boards*, including folders (to share on a networked drive or with services like DropBox), Posit Connect, Amazon S3, Google Cloud Storage, Azure storage, and Microsoft 365 (OneDrive and SharePoint).
 Pins can be automatically versioned, making it straightforward to track changes, re-run analyses on historical data, and undo mistakes.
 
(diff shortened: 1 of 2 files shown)


README.Rmd
```diff
@@ -29,10 +29,6 @@ The pins package publishes data, models, and other R objects, making it easy to
 You can pin objects to a variety of pin *boards*, including folders (to share on a networked drive or with services like DropBox), Posit Connect, Amazon S3, Google Cloud Storage, Azure storage, and Microsoft 365 (OneDrive and SharePoint).
 Pins can be automatically versioned, making it straightforward to track changes, re-run analyses on historical data, and undo mistakes.
 
-pins 1.0.0 includes a new more explicit API and greater support for versioning.
```
juliasilge, line 32: pins 1.0.0 was in Oct 2021 so I think it's time to remove this, to avoid confusion for folks coming to this page afresh. The vignette about moving to pins 1.0.0 is still included.

isabelizimm (approved):
This is exactly what we are looking for! I think the combination of "cannot manage concurrent writes" and "using the app at the same time" hits all the right notes.

PR #836, opened 2024-08-02
Some maintenance on GH actions

juliasilge:
There are two changes here:

- Updating the lock GH action to use a more recent version which hopefully won't fail so much
- Skipping the webfakes tests on the test coverage action, since it currently fails


juliasilge:
Change on the test coverage action is related to https://github.com/r-lib/webfakes/issues/103

PR #854, opened 2025-01-10
Fix tests that use webfakes, in covr

gaborcsardi:


juliasilge (approved):
The [test coverage action now passes](https://github.com/rstudio/pins-r/actions/runs/12711948445/job/35436428296?pr=854)! 🎉

PR #862, opened 2025-04-08
rsconnect::accounts returns a zero-length data.frame and previously returned NULL

### LINKED ISSUE #861: rsconnect::accounts returns a zero-length list, not null
Originally reported in https://github.com/rstudio/rsconnect/issues/1149

https://github.com/rstudio/pins-r/blob/0898a513bb11b1c0c14ddd041fdb38c80a5b982c/R/board_connect_server.R#L69-L72

The `rsconnect::accounts()` function returns a zero-length `data.frame` when there are no known accounts.

https://github.com/rstudio/rsconnect/blob/2ac4925a96520254551797adc64e3762a3a1ab1a/R/accounts.R#L23-L32

This changed with the rsconnect work in https://github.com/rstudio/rsconnect/pull/682, which was included in the rsconnect-1.0.0 release. Versions prior to rsconnect-1.0.0 return NULL when there were no accounts.

aronatkins:
fixes #861

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 * Prepends 'https://' to Databricks host if missing (#855).
 
+* Fixed handling of `rsconnect::accounts()` results (#861).
+
 # pins 1.4.0
 
 ## Lifecycle changes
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
Thank you so much, @aronatkins! 🙌

Issue #865, opened 2025-05-08
qs2 format


igordot:
[qs2](https://github.com/qsbase/qs2) is the successor to the qs package. Should that be an option as well? Or maybe even a replacement?

juliasilge:
Thanks for this suggestion! 🙌 We can use this issue to evaluate interest in supporting qs2 in pins.

juliasilge:
Closed by #866

PR #866, opened 2025-05-13
Add support of the `qs2` format

atsyplenkov:
Hey,

This PR introduces the `qs2` format, the successor to the `qs`, as proposed in #865. In a nutshell, it adds:
* a new family of functions, `read_qs2()` and `write_qs2()`
* updated documentation and tests
* a deprecation warning for the `qs` format.

I am unsure about the latter, as the lifecycle of the `qs` package is unclear to me. However, it will certainly be deprecated in the foreseeable future; the only unknown is when. See https://github.com/qsbase/qs/issues/103

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,6 +58,7 @@ Suggests:
     openssl,
     paws.storage,
     qs,
+    qs2,
     R.utils,
     rmarkdown,
     rsconnect,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pins (development version)
 
+* Added support of the `qs2` format (#865)
+
 # pins 1.4.1
 
 * Support new `preview_data` parameter for pin previews on Posit Connect (#850).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pins (development version)
 
-* Added support of the `qs2` format (#865)
+* Added support of the `qs2` format (#865, @atsyplenkov).
 
 # pins 1.4.1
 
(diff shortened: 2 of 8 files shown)


atsyplenkov:
Alternatively, for back compability, we can implement similar approach to what can be seen in [ropensci/targets:](https://github.com/ropensci/targets/)

https://github.com/ropensci/targets/blob/29b8a726d190d1cf6ec7eb5e2ef4de01cf7b0976/R/class_qs.R#L13-L31

I.e. keep only one type "qs". All new pins will be saved with qs2::qs_save(). For existing pins, we can first try qs2::qs_read() and then fall back on qs::read(). qs2::qs_read() automatically detects the legacy qs format and throws an informative error.

juliasilge:
Thank you so much for this PR! 🙌 

Based on what you have shared, I think I would lean toward _replacing_ qs with qs2 like targets does, rather than adding an additional way to read/write files. Just to confirm, the new package can read a file written by the old package, correct? If so, would you be up for changing this PR to take that approach instead (including support for reading old `.qs` pins)?

atsyplenkov:
> Just to confirm, the new package can read a file written by the old package, correct?

Unfortunately, no. The new package can definitely detect the `.qs` format, but it will throw an error after attempting to read it. So, a possible solution is as follows:
```r
read_qs <- function(path) {
  tryCatch(
    qs2::qs_read(
      file = path
    ),
    error = function(condition) {
      qs::qread(
        file = path
      )
    }
  )
}
```
Nevertheless, I am happy to update the PR with the current approach.

juliasilge (approved):
Thank you so much for this contribution, @atsyplenkov! Since the qs2 package can't read the old qs files, let's stick with your original approach.

PR #871, opened 2025-07-13
Force identical write

lbm364dl:
Before (currently wrong at https://pins.rstudio.com/articles/pins.html#metadata):
image

After:
image

### DIFF
--- a/vignettes/pins.Rmd
+++ b/vignettes/pins.Rmd
@@ -122,7 +122,7 @@ board %>% pin_write(mtcars,
   metadata = list(
     source = "Henderson and Velleman (1981), Building multiple regression models interactively. Biometrics, 37, 391–411."
   ),
-  # Necessary if only changing metadata but file content is the same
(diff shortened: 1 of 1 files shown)


vignettes/pins.Rmd
```diff
@@ -121,7 +121,9 @@ board %>% pin_write(mtcars,
   description = "Data extracted from the 1974 Motor Trend US magazine, and comprises fuel consumption and 10 aspects of automobile design and performance for 32 automobiles (1973–74 models).",
   metadata = list(
     source = "Henderson and Velleman (1981), Building multiple regression models interactively. Biometrics, 37, 391–411."
-  )
+  ),
+  # Necessary if only changing metadata but file content is the same
```
juliasilge, line 125:
```suggestion
  # Necessary if only changing pin metadata but pin content is the same
```

juliasilge (approved):
Thank you so much, @lbm364dl! 🙌

PR #877, opened 2025-08-25
Allow writing pins in multiple formats

### LINKED ISSUE #788: Feature request: option to deploy human-readable format in parallel with binary deployment for datasets
A common workflow for me is to deploy a pin containing a dataset to Connect in both a binary and human-readable format. This usually happens when R/python users want a fast, consistently typed binary (like `qs` or `arrow`) and other business users want to be able to download from Connect in `csv` format to pull into a spreadsheet.

Currently, I have to deploy both of these separately, with unique names (and GUIDs). When distributing the pin or a when a user searches for the pin in Connect I either have to inform them of both deployments/names or give them the correct link/name based on their tooling.

For example, my `pin_write()` calls usually look like this:
```
board %>%
  pin_write(account_information, 
            name = "account_information", 
            type = "arrow"
            )

board %>%
  pin_write(account_information,
            name = "account_information_csv", 
            type = "csv"
            )
```

I'm not sure of the best way to implement this in terms of interface, but in my mind's eye I see something like an option to `include_human_readable`:
```
board %>%
  pin_write(account_information, 
            name = "account_information", 
            type = "arrow",
            include_human_readable = TRUE
            )
```

or even an option to include multiple types:
```
board %>%
  pin_write(account_information, 
            name = "account_information", 
            type = c("arrow", "csv")
            )
```

Another benefit of "coupling" multiple deployment types of the same deployment could be having import redundancy in the case that the pin reader does not have the required package to download a special type like `arrow` or `qs`.

Thanks!

lbm364dl:
Closes #788.

Allowing multiple types in `pin_write` required some minor tweaks in `pin_read`. Instead of just choosing an arbitrary type from those available to read from, I also added a `type` argument in `pin_read`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * Added download progress bar for `board_url()` (#872, @lbm364dl).
 
+* Added support for writing pins with multiple types, like `type = c("rds", "csv")` (#877, @lbm364dl).
+
 # pins 1.4.1
 
 * Support new `preview_data` parameter for pin previews on Posit Connect (#850).


lbm364dl:
> Was there more you wanted to do here before a review, @lbm364dl? I notice it is still draft.

Yes, my intention was to address this thing I mentioned in the linked issue:

> If I have larger files, ideally I wouldn't want users to download a CSV they're not even going to use, because they would just use the other compressed/fast format

I didn't have time to look into that. But it should be independent anyway. I can try in another PR. What is present here should already be enough to close the issue.

A review is more than welcome.

PR #884, opened 2025-10-28
ci: rewrite Connect integration tests using `with-connect` Action

### LINKED ISSUE #883: ci: rewrite Connect integration tests using `with-connect` Action
See https://github.com/posit-dev/connectapi/pull/475 for a proof of concept using this new GitHub Action, which encapsulates the logic for spinning up a test Connect server with Docker. Jotting down some notes here as I read through the existing tests:

* The current setup creates 4 users but only uses 2 of them, and the second user is only used to verify behavior that Connect itself tests (that collaborators are authorized to edit content). So we can cover all of the relevant behavior for pins using the single user account that `with-connect` will bootstrap.
* Test helpers should be moved out of the package proper and into setup.R (though most of them can probably be deleted with this change).
* Is it useful to preserve the PTD option if the local test Connect server isn't running?

nealrichardson:
https://github.com/posit-dev/with-connect

Simplifies the test setup a bit too, and adds in a build matrix to test against multiple versions of Connect, including as far back as 2022.10.0 (the oldest that `with-connect` supports) and including a pointer to whatever the latest release is.

Fixes #883

### DIFF
--- a/R/board_connect.R
+++ b/R/board_connect.R
@@ -733,18 +733,19 @@ rsc_v1 <- function(...) {
 
 # Testing setup -----------------------------------------------------------
 
-# Connect testing: prefers using CONNECT_SERVER and CONNECT_API_KEY env vars,
-# but can fall back to Posit's demo PTD server if you've logged in with rsconnect
+# Connect testing: prefers using Posit's demo PTD server (if you've logged in
+# with rsconnect), else checks CONNECT_SERVER and CONNECT_API_KEY env vars,
+# and if neither are available, skips the tests.
 board_connect_test <- function(...) {
-  if (nzchar(Sys.getenv("CONNECT_API_KEY"))) {
-    board_connect(auth = "envvar", ...)
-  } else if (connect_has_ptd()) {
+  if (connect_has_ptd()) {
     board_connect(
       ...,
       server = "pub.demo.posit.team",
       auth = "rsconnect",
       cache = fs::file_temp()
     )
+  } else if (nzchar(Sys.getenv("CONNECT_API_KEY"))) {
+    board_connect(auth = "envvar", ...)
   } else {
     testthat::skip(
       "board_connect_test() requires CONNECT_API_KEY or Posit's demo PTD server"
--- a/R/board_connect_url.R
+++ b/R/board_connect_url.R
@@ -70,13 +70,13 @@ vanity_url_test <- function(env = parent.frame()) {
 }
 
 board_connect_url_test <- function(...) {
-  if (nzchar(Sys.getenv("CONNECT_API_KEY"))) {
+  if (connect_has_ptd()) {
+    board_connect_url(..., cache = fs::file_temp())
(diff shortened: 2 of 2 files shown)


tests/testthat/setup.R
```diff
@@ -1,3 +1,31 @@
...
+      auth = "rsconnect",
+      cache = fs::file_temp()
+    )
+  } else {
+    testthat::skip(
+      "board_connect_test() requires CONNECT_API_KEY or Posit's demo PTD server"
+    )
+  }
+}
+
```
juliasilge: Can you move the code here in this file back to where it was in the package itself? IMO this would be a loss in terms of package dev experience, and is not consistent with all the other boards.

R/board_connect.R
```diff
@@ -79,15 +79,16 @@
...
-board_connect <- function(auth = c("auto", "manual", "envvar", "rsconnect"),
-                          server = NULL,
-                          account = NULL,
-                          key = NULL,
-                          cache = NULL,
-                          name = "posit-connect",
-                          versioned = TRUE,
-                          use_cache_on_failure = is_interactive()) {
-
+board_connect <- function(
```
nealrichardson: Sorry for the reformatting, this was just format on save with `air`. I can back this out if you want a tighter diff. Can also make a separate PR to `air` everything if you'd like.

nealrichardson: https://github.com/rstudio/pins-r/pull/885, if you want we could merge that first and I can rebase out these here.

R/board_connect.R
```diff
@@ -733,15 +733,25 @@ rsc_v1 <- function(...) {
...
 # Testing setup -----------------------------------------------------------
 
+# Connect testing: prefers using CONNECT_SERVER and CONNECT_API_KEY env vars,
+# but can fall back to Posit's demo PTD server if you've logged in with rsconnect
 board_connect_test <- function(...) {
-  if (connect_has_ptd()) {
-    board_connect_ptd(...)
+  if (nzchar(Sys.getenv("CONNECT_API_KEY"))) {
+    board_connect(auth = "envvar", ...)
+  } else if (connect_has_ptd()) {
```
juliasilge: Thank you for working on this! Can you switch this back to the original order so we check for `connect_has_ptd()` first, for the local testing experience using rsconnect (vs. API keys)?

juliasilge (changes requested):
This looks like a really nice improvement for our CI!

R/board_connect.R
nealrichardson: Done in https://github.com/rstudio/pins-r/pull/884/commits/7ecbc48060b1fcf560f431858854d17731d321a2

nealrichardson:
> Thank you so much! 🙌
> 
> It would be awesome to get similar changes applied to the Python package when you or someone else get a chance.

Yup, headed there next! Just wanted to do this one first since I had deleted the connectapi helpers.

PR #885, opened 2025-10-28
chore: Reformat with air

nealrichardson:
Do it all at once so that every time we touch files in Positron we don't create extra diffs.

### DIFF
--- a/R/board.R
+++ b/R/board.R
@@ -124,7 +124,7 @@ board_cache_path <- function(name) {
   # R_CONFIG_ACTIVE suggests we're in a production environment
   if (has_envvars("R_CONFIG_ACTIVE") || has_envvars("PINS_USE_CACHE")) {
     path <- fs::dir_create(tempdir(), "pins")
-  } else if (has_envvars("PINS_CACHE_DIR") ) {
+  } else if (has_envvars("PINS_CACHE_DIR")) {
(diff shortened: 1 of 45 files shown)


nealrichardson:
Both CI failures are expected. #884 fixes posit-connect; I filed #886 for the check-boards databricks issue.

juliasilge:
Can you tell me why the failure on the `posit-connect` action is expected?

nealrichardson:
> Can you tell me why the failure on the `posit-connect` action is expected?

The action [`:::`s a `connectapi` function](https://github.com/rstudio/pins-r/blob/main/.github/workflows/posit-connect.yaml#L41-L44) that I [deleted yesterday](https://github.com/posit-dev/connectapi/pull/475) when removing the `HackyConnect` business in favor of the `with-connect` utility, the same one I put up a PR to add here (#884). I didn't realize it was being called anywhere outside of `connectapi`'s own tests. 

The existing workflow would probably pass if https://github.com/rstudio/pins-r/blob/main/.github/workflows/posit-connect.yaml#L30 were pinned to a particular release of `connectapi` or a github SHA before yesterday. But #884 also fixes it--the (new) `posit-connect` workflow is passing on that PR. I put this PR up now so that, assuming you wanted to `air` out the codebase anyway, we could merge this first and then I could rebase #884 so it only had the relevant changes. But it doesn't have to go in that order.

PR #898, opened 2026-03-13
Switch pins to use parquet by default for dataframes

### LINKED ISSUE #666: Consider changing default format for dataframes to arrow
We have seen users who write using the default from R, and then are frustrated when their Python colleagues can't read. We have considered changing to arrow for a long time:

https://github.com/rstudio/pins-r/blob/3d2dc6b967b01dacee08d46af60aa5981ed774ba/R/pin-read-write.R#L115

Does arrow have enough usage in the community for this to be reasonable? It would be a much better choice if interoperability is one of the main reasons people use pins (to read [with Python](https://rstudio.github.io/pins-python/)).

juliasilge:
Closes https://github.com/rstudio/pins-r/issues/666

I think we have enough information that this is working well for people that we can switch out the default here.


NEWS.md
```diff
@@ -1,4 +1,8 @@
-# pins (development version)
+# pins (development version to be released as 1.5.0)
+
+## Breaking changes
+
+* Changed the default file type for data frames from RDS to Parquet when writing a pin without an explicit `type`. Parquet offers better interoperability with other languages, especially Python (#898).
```
juliasilge, line 6: I'm going to call this a breaking change and bump to 1.5.0.

edavidaja: _harrumphs in semver_

hadley (approved):
🥳

